PR #49, opened 2014-11-17
Java and Csharp generation

ldallolio:
Dear Mr. Paviot,

I needed to generate csharp interfaces for OCE, so I configured pythonocc to do so... I understand that this is a little bit against the name of the project... ;-) I think that you might want to consider to have this in the main fork (I have tested both java and csharp and, at least in simple cases, it works). I will keep enhancing cmake integration etc. but please let me know your opinion about all this and in any case, thank you very much in advance for your work

Best regards,

Luca


tpaviot:
@ldallolio I close this pull request, I cannot test/maintain anything else that pythoncode. For the java/swig wrapper for opencascade, please have a look at http://jcae.sourceforge.net/, your code may be appreciated.

ldallolio:
I perfectly understand, it is probably the best thing to do :-)

Meilleurs voeux,

Luca

PR #69, opened 2015-01-07
fixes a number of issues with the Qt viewer

jf---:
fixes a number of issues with the Qt viewer:

* when Shift-RMB dragging, no rectangle would be overpainted in the viewer. Shift-RMB dragging draws a frame that will be zoomed, so this is really important...
* Shift select was not implemented, which allows for multiple selection

These changes allow for overpainting of the OCC OpenGL viewer. This paves the way for really cool widget to be drawn on top of the OCC OpenGL viewer. OCC has a API for painting a layer on top of the viewer, but its horrible and does not allow for widgets / interactivity.

An example is contributed, "examples/core_visualization_overpaint_viewer.py"

This example is WIP -- when rotating the view, then there's an issue syncing the buffers.
Close to getting there though...

### DIFF
--- a/examples/core_visualization_overpaint_viewer.py
+++ b/examples/core_visualization_overpaint_viewer.py
@@ -1,0 +1,278 @@
+"""
+
+demonstrates overpainting of the OCC OpenGL viewer to share its OpenGL Context with Qt
+
+this allows for interesting interfaces, where elements are drawn as an overlay
+
+:ref:`Catia's structure browser` is a well known application of this idea
+
+the example is extends Qt's :ref:`OpenGL overpainting example`
+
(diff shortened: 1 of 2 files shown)


tpaviot:
@jf--- Interesting, I have to test this (I'm currently using pyside). Here is what should be done:

* fix Shift-RMB only in pyqt4Display.py. I don't want to add a dependency over pyopengl

* in the example, adds a code block to check whether or not pyopen gl is installed. Exit if not

* modify pyside code as well so that both pyqt and pyside based displays are consistent.

jf---:
OK, thanks for the constructive feedback, makes perfect sense, I'm on it.
Definitely I'm behind on pyside support.

tpaviot:
I pushed review/qt-fixes, which merges pyqt4Display and pysideDisplay (only one line differs between these two files). May be it would be better to work on top of this branch. This way, your changes will be available for both pyside/pyqt4 and it will be easier to test.

jf---:
OK, I just saw that, many thanks for the headsup!

jf---:
Moved to PR #73

Issue #70, opened 2015-01-07
TopoDS_Shape subclasses cannot be inherited from


jf---:
The following is a pretty serious regression... 
AFAIK its not possible to inherit any TopoDS_Shape decendents

This is because the TopoDS_Edge ( and TopoDS_Face and friends ) 
do not define a ( copy? ) constructor that takes a existing ( TopoDS_Edge here ) instance as argument

I dont see how this can be by-passed using OCC.TopoDS.topods.Edge or OCC.TopoDS.topods_Edge

That means that we can't inherit from TopoDS_Shape subclasses... 

```python
from OCC.BRepBuilderAPI import BRepBuilderAPI_MakeEdge
from OCC.TopoDS import topods_Edge, TopoDS_Edge
from OCC.gp import gp_Pnt


line = BRepBuilderAPI_MakeEdge(gp_Pnt(), gp_Pnt(100,0,0)).Edge()

class InheritEdge(topods_Edge):
    def __init__(self, edge):
        super(InheritEdge, self).__init__(edge)

"""

cannot inherit from:

Traceback (most recent call last):
  File "/Users/jelleferinga/GIT/pythonocc-core/src/addons/Utils/tst.py", line 8, in <module>
    class InheritEdge(topods_Edge):
TypeError: Error when calling the metaclass bases
    function() argument 1 must be code, not str
"""

class InheritEdge(TopoDS_Edge):
    def __init__(self, edge):
        # no constructor defined... 
        # TopoDS_Edge() just returns an uninitialized TopoDS_Edge...
        super(InheritEdge, self).__init__(edge)
```

jf---:
following the [OCC 6.8.1 docs](http://dev.opencascade.org/doc/refman/html/class_topo_d_s___edge.html#details), the TopoDS_Edge indeed does not define a constructor... 

I need to dig a little further...

jf---:
probably this class should now inherit from TopoDS_Shape 

So, I think initialization is done in a really different way now, [see the docs here](http://dev.opencascade.org/doc/refman/html/class_topo_d_s___builder.html)

Perhaps our implementation could look like this:

```python

class MyFancyEdge(TopoDS_Shape):
    def __init__(self, edge):
        super(InheritEdge, self).__init__()
        # method below not available... its protected
        TopoDS_Builder.MakeShape( self, edge )

```

jf---:
from OCC import BRep_Builder

[...]

PR #71, opened 2015-01-08
Added inherit topods example and unittest

tpaviot:
Implement example and unittest related to issue #70


jf---:
Ah, so cool to see this made into a unittest! 
Great!

PR #73, opened 2015-01-08
review/qt-fixes

jf---:
a fresh PR from feedback of PR #69

PR #74, opened 2015-01-09
Review/qt fixes

tpaviot:
fixes a number of issues with the Qt viewer:
    
    * when Shift-RMB dragging, no rectangle would be overpainted in the viewer. Shift-RMB dragging draws a frame that will be zoomed, so this is really important...
    * Shift select was not implemented, which allows for multiple selection
    
    These changes allow for overpainting of the OCC OpenGL viewer. This paves the way for really cool widget to be drawn on top of the OCC OpenGL viewer. OCC has a API for painting a layer on top of the viewer, but its horrible and does not allow for widgets / interactivity.
    
    An example is contributed, "examples/core_visualization_overpaint_viewer.py"


tpaviot:
Related to issue #73

Issue #98, opened 2015-05-05
Travic-CI broken because of bad oce version


tpaviot:
Current travis-ci builds fail. pythonocc current master requires oce-0.16.x to compile, but the ppa from which oce binaries are pulled recently moved up to 0.17. Thus the compilation fails.

This can be solved using two solutions:

1. the freecad tems restore oce-0.16

2. we (at oce or pythonocc) provide a launchpad ppa for oce releases.

The second solution would be quite better.

jmwright:
I think an OCE PPA that's sync'd with the pythonocc PPA, and that we can be sure won't break when new versions are released (option 2), would be very helpful.

jmwright:
OCE 0.16 is available on [Alioth](http://anonscm.debian.org/cgit/debian-science/packages/oce.git). The [git-buildpackage](https://www.debian.org/doc/manuals/maint-guide/build.en.html) (section 6.5) utility is used to build Debian packages from git.

```
sudo apt-get install quilt libfreetype6-dev tcl8.5-dev tk8.5-dev libgl2ps-dev libfreeimage-dev
git clone https://anonscm.debian.org/git/debian-science/packages/oce.git
cd oce
sudo git-buildpackage --git-ignore-branch
```

In theory, it could just be a matter of placing the generated source packages in a PPA after that. I haven't tried it yet though because the build is taking quite awhile. I can report report back and keep pursuing this if it seems like an acceptable option.

tpaviot:
I didn't know about git-buildpackage, thank you for pointing this.

Launchpad has a farm build system, that prevents from compiling deb packages on your machine and upload them to launchpad. I already did it for pythonocc-core, I could do the same for oce, it's just that it's not straightforward and a good knowledge of debian packaging is required. It was quite a good think the freecad people take care of that.

PR #100, opened 2015-05-07
Fix travis build

tpaviot:
Previous travis build was configured to use OCE binaries from the free-cad launchpad ppa. Unfortunately, after the upgrade of freecad up to oce-0.17, their ppa is not sync'ed anymore with pythonocc-core (which requires oce-0.16).

In ths new travis file, oce-0.16.1 is downloaded/compiled/installed before pythonocc compilation. This takes some precious time, leading to give up with gcc tests (only clang left).

This however should be enough to have travis-ci fixed.

### DIFF
--- a/.travis.build-oce.sh
+++ b/.travis.build-oce.sh
@@ -1,0 +1,28 @@
+#! /bin/sh
+set -e
+
+
+#  Default values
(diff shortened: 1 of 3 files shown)


tpaviot:
This branch fixes issue #98

Issue #104, opened 2015-05-20
classic bottle example missing


jf---:

jmwright:
I would be happy to submit my bottle example for review if it will help with this. I wrote it using pythonocc-core 0.16. It needs improvement, but it works and was adapted from older examples.

jf---:
Great, could you submit a PR?
Only comment is that I'd make casting the Geom_Plane from a face into a function.

Thanks @jmwright !

```python
myBody_traverser = Topo(myBody)
for aFace in myBody_traverser.faces():
    if face_is_plane(aFace):
        aPlane = Handle_Geom_Plane_downcast(BRep_Tool_surface(aFace)).GetObject()
        aPnt = aPlane.Location()
        aZ = aPnt.X()
        if aZ>zMax:
            zMax = aZ
            faceToRemove = aFace
```

```python

def geom_plane_from_face(aFace):
    """ returns the geometric plance entity from a planar surface"""
    return Handle_Geom_Plane_downcast(BRep_Tool_surface(aFace)).GetObject()

myBody_traverser = Topo(myBody)
for aFace in myBody_traverser.faces():
    if face_is_plane(aFace):
        aPlane = geom_plane_from_face(aFace)
        aPnt = aPlane.Location()
        aZ = aPnt.X()
        if aZ>zMax:
            zMax = aZ
            faceToRemove = aFace
```

jmwright:
@jf--- I made the change you suggested, but I am unable to find the equivalent of Topo in pythonocc-core. Therefore, I have to use a bulky while loop instead of iterating using a for loop. That's one of the things that I'm thinking can be improved.

jf---:
Cool! Code looks really clean & mint, thanks Jeremy...

@jmwright , OK, let me add the topology.py module, its a really key piece of functionality.
@tpaviot , what about src/addons/Utils/topology.py?

that PR would rework also the examples to use that module.
@jmwright , are you OK to wait for this PR to be merged post merging topology.py?

jmwright:
@jf--- Thanks. Glad the code looks good. I'm ok waiting for my PR to be merged. I had actually expected it to take awhile while the example was improved (with things like the inclusion of topology.py).

tpaviot:
Fixed by PR #105

PR #105, opened 2015-05-20
Added a classic OCC bottle example.

jmwright:
This is in response to issue https://github.com/tpaviot/pythonocc-core/issues/104
Based on the official [OCC bottle tutorial](http://www.opencascade.org/doc/occt-6.7.1/overview/html/occt__tutorial.html) and the example attached [here](https://mail.gna.org/public/pythonocc-users/2011-05/msg00013.html).

This example needs some work. For example, OCC.Utils.Topology.Topo seems to be non-existent in pythonocc-core, and so a Python for loop isn't used to iterate over faces. A bulky while loop is used instead. The example runs on an Ubuntu 14.04 64-bit VM running a OCE/pythonocc-core 0.16 setup via [this script](https://github.com/jmwright/pythonocc_oce_setup).

### DIFF
--- a/examples/core_classic_occ_bottle.py
+++ b/examples/core_classic_occ_bottle.py
@@ -1,0 +1,201 @@
+#!/usr/bin/env python
+
+import math
+
+from OCC.gp import gp_Pnt, gp_OX, gp_Vec, gp_Trsf, gp_DZ, gp_Ax2, gp_Ax3, gp_Pnt2d, gp_Dir2d, gp_Ax2d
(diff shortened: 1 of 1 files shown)


jmwright:
Looks like this failed with Travis because the test run exceeded 50 minutes.

tpaviot:
Thanks @jmwright for this contribution.

jf---:
congrats on your 1st merge @jmwright :8ball:

jmwright:
Thanks!

PR #110, opened 2015-06-18
jf/refactor_GC

jf---:
optimizations for the GC

### DIFF
--- a/src/GarbageCollector.py
+++ b/src/GarbageCollector.py
@@ -16,31 +16,24 @@
 ##along with pythonOCC.  If not, see <http://www.gnu.org/licenses/>.
 
 # Init logging system
-import logging
-logging.basicConfig()
-logger = logging.getLogger('GarbageCollector')
-
-
-def set_debug():
-    logger.setLevel(logging.DEBUG)
-
+from collections import deque
 
 class GarbageCollector(object):
     """ Garbage collector for OCC objects
     """
     def __init__(self):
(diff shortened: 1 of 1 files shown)


tpaviot:
@jf--- I fixed travis build syste. Is it possible that you rebase your branch on top of master to benefit from this fix ?

Interesting contribution, is there any way to benchmark the benefits in terms of computation time ?

jf---:
@tpaviot , thanks, yeah, so rebased...

actually, this is hard to profile.

let me cite from [stackoverflow](http://stackoverflow.com/questions/7149283/implementation-of-dequeue-in-python-using-linkedlist):

> List append performance is hit and miss because it uses realloc() under the hood. As a result, it tends to have over-optimistic timings in simple code (because the realloc doesn't have to move data) and really slow timings in real code (because fragmentation forces realloc to move all the data). In contrast, deque append performance is consistent because it never reallocs and never moves data.

i noticed serious performance degradation with long running pyocc programs, especially when generator a lot of instances ( like gp_Vec when doing optimization, millions and millions of them ).

the application seems more repsonisve now under those conditions...
again, doing serious comparison would be pretty challenging.
I *think* this works quite a bit better, but its also a case of ruling things out, in terms of performance.
in pure python, this *could* be a lot more performant, to which extent is hard to tell...

jf---:
@tpaviot , cool wrt to the Travis fix btw, nice work!

PR #112, opened 2015-06-19
Removed Visual3d_Layer::DrawText method that can't be wrapped on Win3…

tpaviot:
…2/MinGw


jf---:
at some point this method is not that important... I realize it was a feature request from my side...
but since I've used Qt's overpainting ( in the examples ) which clearly is a much cleaner / more functional approach

PR #123, opened 2015-07-28
port of the good ol' Topology module and unittest too challenging and central to have to re-invent...

jf---:
port of the good ol' Topology module and unittest
too challenging and central to have to re-invent...


jf---:
@tpaviot, perhaps this commit will spark some ideological discussion:

IMHO we need to supply a set of central modules that provide core functionality to pythonocc.
I think in the current state we've gone a little overboard: a *lot* of knowledge and experience has gone into developing these modules. So if they're really central concepts, that have a rigorous, well documented, well tested set of functionality, we should accept that into pythonocc-core.

A minimal set is needed; for instance, when porting now OCC.DataExchange.STEP.STEPExport, how will you test the resulting STEP file? You need OCC.DataExchange.STEP.STEPImport to test OCC.DataExchange.STEP.STEPExport, and OCC.Utils.Topology to perform a meaningful test.

More adventurous / non-central / not well tested / additional dependencies requiring, more volatile code indeed should move to pythonocc-contrib. Clearly we've been too permissive in the previous repo. Lets strike the right balance :sunglasses:

tpaviot:
@jf--- Old Utils components development/port is now hosted at pythonocc-utils (https://github.com/tpaviot/pythonocc-utils). Please submit there any improvement/change to any of these  modules. Open an issue on the pythonocc-utils tracker to discuss the opportunity to move one of this modules to pythonocc-core.

jf---:
I feel this is a bad idea...

I think pythonocc needs a nr of essential modules such as helping you to read / write STEP & IGES files and to be able to explore the topology of a BRep. This is core functionality, that has been part of pythonocc for 7 years :sunglasses:. 

While perhaps in terms of software engineering, there might be something to be said to keep a strict separation, I think its highly unpractical. Users should not have to install 2 modules to get basic functionality going. This is complex code, and complete covered in unittest. 

I should add a proper example on how to use the module though, true...

PR #132, opened 2015-08-01
menu's to be embedded in the viewer

### LINKED ISSUE #131: proposed fix, for issue with SimpleGui's menus in recent pyqt versions
I upgraded Anaconda's pyqt version recently, and noticed that menu's didnt work out well.
First, the menu isnt visible until you hide all other apps
Second, I much prefer having the menu embedded in the viewer.

Both can be accomplished by the proposed fix.
Tested on both pyqt and pyside.

fix for the code [found here](https://github.com/tpaviot/pythonocc-core/blob/master/src/addons/Display/SimpleGui.py#L119):

current implementation:
```python

                if not sys.platform == 'darwin':
                    self.menu_bar = self.menuBar()
                else:
                    # create a parentless menubar
                    # see: http://stackoverflow.com/questions/11375176/qmenubar-and-qmenu-doesnt-show-in-mac-os-x?lq=1
                    # noticeable is that the menu ( alas ) is created in the topleft of the screen, just
                    # next to the apple icon
                    # still does ugly things like showing the "Python" menu in bold
                    self.menu_bar = QtGui.QMenuBar()
                self._menus = {}

```

proposed fix
```python
                if  sys.platform == 'darwin':
                    QtGui.qt_mac_set_native_menubar(False)
                self.menu_bar = self.menuBar()
```

jf---:
resolves issue with recent pyqt versions, where menu's are no longer available
fixes #131

### DIFF
--- a/src/addons/Display/SimpleGui.py
+++ b/src/addons/Display/SimpleGui.py
@@ -18,6 +18,7 @@
 ##along with pythonOCC.  If not, see <http://www.gnu.org/licenses/>.
 
 import sys
+
 from OCC import VERSION
 
 
@@ -33,11 +34,13 @@ def get_backend():
         return 'qt-pyqt4'
     except:
         pass
+
     try:
         from PySide import QtCore, QtGui
         return 'qt-pyside'
     except:
         pass
+
     # Check wxPython
     try:
         import wx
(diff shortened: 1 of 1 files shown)


tpaviot:
Is this change backward compliant with older pyqt releases? (I can't test for the moment)

PR #134, opened 2015-08-01
working on ideas for the new example structure

jf---:
here are a few ideas, based on the mesh-from-shape example

using Readme.md's allow to easily render images of the examples in GitHub
question is whether it would be a good idea to move towards such a folder layout

now, we have a lot of core_* examples, where I think it makes sense to have a folder for the many core_display examples for instance. the cool thing here, is that those folder can be considered modules.

if all examples, start by clearing the viewer, than its easy to create a "run_all_example.py", where the folders becomes menu's, and all function in the python files are then added to the menu. I think that would be pretty attractive, since it allows to go through much more material, without having to reboot python that often.

the update in the example is that the resulting mesh is shifted next to the input shape, which makes it a clearer example


tpaviot:
That's definitely what is needed. Todo:

* be able to automagically generate the readme and screenshot from the example and code comments (this would ensure to keep sync'ed with examples changes)

* provide a draft for the root example

* provide a kind of "readme template" to make sure all readmes are consistent.

I'm ok to move to this new structure as soon as these 3 points are completed.

tpaviot:
Closing this branche, it did not move from the the last comment.

PR #135, opened 2015-08-01
inform the user that you need gl2ps for this example to work out

jf---:


tpaviot:
That's oce, not pythonocc, that needs to be built with gl2ps. Rather than a single comment in the source, maybe a print () would be better?

PR #145, opened 2015-08-27
Moved conda pkg WIP from pythonocc-core to pythonocc-contrib

tpaviot:
The conda pkg work is OSX specific so far. This contribution is moved to pythonocc-contrib, where it can get more mature.


jf---:
Agreed. Thanks for moving to contrib...

PR #154, opened 2015-12-02
Fix windows compilation

rainman110:
Compiling pythonocc-core on windows currently failed due to 2 reasons:
 1.  Non-exported classes beeing wrapped by swig
 2.  Failure during ```make install``` because of a duplicated install prefix

This pull request should fix both problems.


tpaviot:
Thanks for sharing. This pull requests hould be splitted into 2 differents commit. The install path is related to pythonocc-core. The second one (non-exported classes beeing wrapped by swig) is related to the generation rules of the wrapper and should be submitted to pythonocc-generator (https://github.com/tpaviot/pythonocc-generator).

tpaviot:
BTW, which compiler to you use on Windows ?

rainman110:
I understand. Then we'd have to modify the generator tool, that it detects, if STANDARD_EXPORT is set. Have you an idea, how to do it? You're probably more into the code of the generator than me.

> BTW, which compiler to you use on Windows ?

For python 2.7 the vc 2008 compiler is recommended. Therefore I use the vs 2008 express edition.

tpaviot:
That should not be that hard. Rules for generating the wrapper are in https://github.com/tpaviot/pythonocc-generator/blob/master/src/generate_wrapper.py

One can add a rule "if not STANDARD_EXPORT in function prototype then do not generate the wrapper"

With cmake, you can use the compiler yuo want, there's no need to use msvc2008 (it's only if you work with distutils).

rainman110:
> With cmake, you can use the compiler yuo want, there's no need to use msvc2008 (it's only if you work with distutils).

Yes, this is true. I normally use msvc 2010 and msvc 2013. But it is recommended to use 2008 in order to avoid mixing different msvc runtimes, which can lead to problems. (http://stackoverflow.com/a/29910249)

rainman110:
I analyzed the OCE headers with the wrapper_generator.py. 

Strangely, there are many methods, that are not exported with ```Standard_Export ``` (for example allmost all methods in TopoDS_Shape). 

If I inspect ```TKBRep.dll``` with Dependency Walker, only ```TopoDS_Shape::HashCode``` is exported (as expected, as it is the only method marked with Standard_Export). Still, all other methods can be wrapper without error. 

Does anyone have an explanation for this?

rainman110:
> Does anyone have an explanation for this?

Okay, figured it out. Many functions are inlined in lxx files. Therefore they don't need to be exported.

The wrapper generator must now detect, if a method is either
 1) Exported with Standard_Export
 2) Or inlined in a lxx file

This makes the wrapping process much more complicated (I think). @tpaviot Do you have an idea, how we can deal with this?

rainman110:
@tpaviot I have rewritten the generator to incorporate inlined methods from lxx files. Now things are getting more and more complex:

 Sometimes, header files include an lxx file that e.g. implement an array. In order to make it work, OCCT does not use polymorphism but using the macros define. A good example is [TopoDS_ListIteratorOfListOfShape.hxx](https://github.com/tpaviot/oce/blob/master/inc/TopoDS_ListIteratorOfListOfShape.hxx). 

Here, the "base class" is TCollection_ListIterator, included in TCollection_ListIterator.lxx. The functionionality is done using the defines at the botton of the hxx file.

This makes the detection of already inlined functions almost impossible. We'd require a proper c++ preprocessor, to resolve things.

For now I think it is easier to have everything as it is and fix the few non-exported methods by hand. @tpaviot: What do you think?

tpaviot:
@rainman110 agree, general rules are sometimes difficult to state. That's actually what I did (remove functions/methods that are not correctly exported), but not "by hand". The list of classes/method to exclude from the wrapper is defined in Modules.py of the pythonocc-generator projet (see the OCE_MODULES list https://github.com/tpaviot/pythonocc-generator/blob/master/src/Modules.py#L124).

rainman110:
@tpaviot I made a pull request on the code generator.

 - Should a make a separate PR for the cmake fix?

 - When do you include the generated (fixed) code in the repo? I could make another PR using the generated code.

Issue #93, opened 2015-03-21
Linker errors for non-exported OCE classes


opticode:
On Windows, using VS 2013 64bit, with pythonocc-core 0.16 (and OCE 0.16) I get around 20 linker errors, e.g.:

    error LNK2019: unresolved external symbol "public: __cdecl BRepMesh_CircleInspector::BRepMesh_CircleInspector(double,int,class ...
    ... referenced in function _wrap_new_BRepMesh_CircleInspector ...

On inspecting these classes, I can see that they only appear in .hxx files in the "src" directory of OCE and are *not* exported, e.g. notice the missing ```Standard_EXPORT``` from ```OCE/src/BRepMesh/BRepMesh_CircleInspector.hxx```:

    //! Implementation of inspection method
    NCollection_CellFilter_Action Inspect (const Standard_Integer theTarget); 

Should these be wrapped at all? The affected classes are:

* BRepMesh_CircleInspector
* BRepMesh_VertexInspector
* BRepMesh_Delaun
* ShapeAnalysis_BoxBndTreeSelector
* Storage_Bucket
* Storage_BucketOfPersistent
* Storage_BucketIterator

In addition, I also get this error:

    error C1091: compiler limit: string exceeds 65535 bytes in length - \Dev\pythonocc-core-0.16.0\Win64\bin\V3dPYTHON_wrap.cxx

tpaviot:
@opticode You're right, these classes/method should not be wrapped.

PR #157, opened 2015-12-03
Code with the windows fixes from the wrapper_generator

rainman110:
The code from this commit is generated using the fixed wrapper_generator (https://github.com/rainman110/pythonocc-generator/tree/ms/fix_windows_build). It should fix all linker errors that occur on windows.

Fixes issue #93


tpaviot:
Everything is ok and consistent with pythonocc-generator changes.

PR #165, opened 2015-12-11
Some work on Tesselator

rainman110:
- Computation of edges
- Interface to Vertices, Normals, and Triangles

### DIFF
--- a/src/Visualization/Tesselator.cpp
+++ b/src/Visualization/Tesselator.cpp
@@ -25,9 +25,13 @@
 #include <BRepMesh.hxx>
 #include <Poly_Connect.hxx>
 #include <Poly_Triangulation.hxx>
+#include <Poly_PolygonOnTriangulation.hxx>
 #include <TColgp_Array1OfPnt.hxx>
 #include <TColgp_Array1OfDir.hxx>
 #include <TColgp_Array1OfPnt2d.hxx>
+#include <TopTools_ListOfShape.hxx>
+#include <TopTools_IndexedDataMapOfShapeListOfShape.hxx>
+#include <TopExp.hxx>
 #include <BRepTools.hxx>
 #include <BRepBndLib.hxx>
 //---------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


jf---:
Cool!
An interface to Edges could be useful too, right?
I think now in python you'd have to do lots of iteration to get all edges?

The ipython work is exciting...

rainman110:
> I think now in python you'd have to do lots of iteration to get all edges?
Unfortunately yes. It would be better, if the swig interface already returns the whole array instead of iterating over each vertex.

The workflow is as follows:
 - Tesselate
 - Iterate over vertices of faces and edges
 - Put all vertices in dictionaries
 - Transmit dictionaries to javascript using the ipython bridge
 - In javascript: put everything into three.js objects (again for loops)

This might be the reason, why my code is quite slow. There is probably some room for optimization though.

jf---:
Ah, thanks for the insight @rainman110 

To make this fast, probably we could create a .json string in pythonocc, where iteration through the mesh is handled in C++ and the .json returned through the SWIG interface.

But this is a nice step fwd!

rainman110:
Good idea! Should be much faster.

rainman110:
There seem to be build problems with the Tesselator. I will have a look into it.

src/Visualization/Tesselator.cpp
```diff
@@ -199,11 +215,9 @@ void Tesselator::Tesselate()
             validFaceTriCount++;
           }
         }
-        this_face->number_of_triangles = validFaceTriCount;
-        facelist.push_back(this_face);
```
rainman110: Deleting these lines was by accident, sorry.

tpaviot:
Greta, interface to vertices/faces/edges was something requested for a while.

We should add an example now.

rainman110:
What about an example, where we plot a shape with matplotlib? There we can explicitly use the tesselator to get the triangles and edges.

jf---:
Initially the tesselator was put in place for generating mesh data to be rendered in webgl.
I think keeping the context of webgl or opengl is where this functionality is most applicable.

tpaviot:
The Tesselator can be used to solve any problem. So far actually the Tesselator is mostly used to render a shape in the browser, but if it can be used for something else, it's cool too. I use to work with numpy/matplolib, an example demonstrating a use of pythonocc/matplotlib would be great IMHO.

rainman110:
I just added a small matplotlib example for the tesselator. I think this is more instructive than a huge opengl/webgl code. See my PR #167

jf---:
@rainman110 , I just saw your PR...
Yes, the MPL example is really cool... sorry I didnt realize this was a good idea before...

PR #167, opened 2015-12-14
Added an example, how to the Tesselator interface

rainman110:
The example plots a shape with matplotlib (3d). In addition to
the triangular surface, also the edges of the shape are drawn.

### DIFF
--- a/examples/core_matplotlib_box.py
+++ b/examples/core_matplotlib_box.py
@@ -1,0 +1,70 @@
+#!/usr/bin/env python
+
+##Copyright 2015 Martin Siggel (martinsiggel@gmail.com)
+##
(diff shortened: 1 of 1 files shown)


rainman110:
This is the output of the draw-to-matplotlib example:

draw_to_mpl

tpaviot:
that's great

KeyDaar:
Hello Martin,

I have installed latest pythonOcc 0.17.2 and trying to run this script but getting error as follows,
`edge_count = tessalator_.ObjGetEdgeCount()`
`AttributeError: 'Tesselator' object has no attribute 'ObjGetEdgeCount'`
Please suggest.
Thank-you in advance.

PR #172, opened 2015-12-23
jf/refactor_overpaint

jf---:
overpaint example fails to overpaint when re-orienting ( panning, rotating ) the view 

fixes #171

Issue #171, opened 2015-12-23
overpaint example fails to overpaint when re-orienting ( panning, rotating ) the view


jf---:
So the following actions invoke the OpenGL window to be re-drawn:

ON_ZOOM 
ON_ZOOM_AREA 
ON_ZOOM_FACTOR 
ON_ZOOM_FITALL 
ON_DYN_ZOOM 
ON_DYN_ROT 
ON_DYN_PAN 
ON_MOVE_TO 
ON_SHIFT_SELECT 
ON_SELECT 
ON_SELECT_AREA 

Now, when you perform any of these commands, say you're zooming or panning, this will interfere with the overpainting on top of the OpenGL window.

So, there's a lack of coordination between actions that invoke the OpenGL window to be redrawn.
The way things work in Qt is that when you call the __update__ method, that all OpenGL actions are performed in the correct order, that is, overpainting the view comes last.

So rather than having the action performed directly ( eg, unorchestrated ), the command is registered as the current action and all redrawing is performed with the __paintEvent__ method, following the Qt idiom.

I guess its best to provide an example: here's  how I've re-implemented the overpaint examples __mouseMoveEvent__:

```python
    def mouseMoveEvent(self, event):
        print("mouse move")
        self.point_on_mouse_move = event

        buttons = int(event.buttons())
        modifiers = event.modifiers()

        # rotate
        if buttons == QtCore.Qt.LeftButton and not modifiers == QtCore.Qt.ShiftModifier:
            self.current_action = ON_DYN_ROT

        # dynamic zoom
        elif buttons == QtCore.Qt.RightButton and not modifiers == QtCore.Qt.ShiftModifier:
            self.current_action = ON_DYN_ZOOM

        # dynamic panning
        elif buttons == QtCore.Qt.MidButton:
            self.current_action = ON_DYN_PAN

        # zoom window, overpaints rectangle
        elif buttons == QtCore.Qt.RightButton and modifiers == QtCore.Qt.ShiftModifier:
            self.current_action = ON_ZOOM_AREA

        # select area
        elif buttons == QtCore.Qt.LeftButton and modifiers == QtCore.Qt.ShiftModifier:
            self.current_action = ON_SELECT_AREA
       
       # -> this invoked self.paintEvent, following the Qt idiom
        self.update()
```

jf---:
This work is in anticipation of moving towards Qt5, with Qt5.6 LTS support being announced and Qt 4.8 support fading 

[...]

PR #175, opened 2015-12-23
jf/pyqt5

### LINKED ISSUE #174: PyQt5 support
with Qt4 reaching no longer being supported [post 19th of Dec 2015])(http://blog.qt.io/blog/2014/11/27/qt-4-8-x-support-to-be-extended-for-another-year/), its time to get going on providing support for Qt5 / PyQt5. 

finally, with Qt5, QML is really becoming to way to do GUI's in Qt, so this wedges into supporting some #173 / QML examples

jf---:
lack of PyQt5 support
fixes #174


jf---:
overlaps with #171

tpaviot:
Can you please rebase onto master after PR #172 was merged. Take care to not overwrite the pep8 changes I made to core_visuaization_over_painter.

I did not test yet the PR, but :+1: 

* changes to SimpleGui and over_painter seems ok to me

* I disagree with the __init__ changes. These changes are Qt specific whereas the display can use wx, ThreeJs and x3dom. There's no need to check for PyQt4, Qt5 or Side if we need to render to x3dom. Leave the __init__ empty as it was is surely the best way.

jf---:
Thanks for the hints @tpaviot ...

Wrt to the "\__init__.py" changes; my thinking was that getting a backend is tangential to loading a module, so the logic / technicalities for loading this module for how to load this module should be moved to "\__init__.py".

So you're right in the sense that the rest of the logic ( wx, x3dom ) should go there too... all those technicalities need to be handled in one single place, that's absolutely true...

I think "\__init__.py" is the right place to do so. If you don't like that idea, then what about creating a backend.py file?

Wrt pep8: I tend to rely on the formatting fixes of PyCharm, but it seems more aggressive measures for pep8 compliance is important. I guess running [autopep8](https://pypi.python.org/pypi/autopep8) is the way to go :sunglasses:

jf---:
OK, this is starting the shape up...

Tested on all qt backends ( -> **core_visualization_overpaint_viewer.py**, more work needs to be done before merging this... to make this PR compatible with the rest of the code base )

What's cool is that handling loading the backend is now well documented & yields informative exceptions. This results in more robust behaviour and a clear separation of concerns.

One thing that isnt too good is the following, from the **core_visualization_overpaint_viewer.py** example:

```python
from OCC.Display.backend import get_backend, get_qt_modules
backend = get_backend()
QtCore, QtGui, QtWidgets, QtOpenGL = get_qt_modules()
from OCC.Display.qtDisplay import qtViewer3d
```

You need to call __get_backend__ before being able to use **from OCC.Display.qtDisplay import qtViewer3d**. It will give you a very informative error :tropical_drink: exception...

Right now, I can't think of a more elegant way to handle this.

The code in **display.py** isnt very DRY... I've explored a more magical method, but mehhh... that become too difficult / magical.... I prefer ladder logic over magics anytime :mushroom: 

Curious to hear what you think...

jf---:
YIKES... 
The import logic is not correct when no backend is specified and PyQt5 is not found...

jf---:
Fixed before mentioned issue...
Things are starting to shape up...

tpaviot:
Started the review. The patch is not py3k compliant. Rather use absolute imports rather than relative e.g.
```python
from OCC.Display import bakcned
```
rather than
```python
import backend
```

Take also care to the statement
```python
except Exception,e
```
which is py2 specific

tpaviot:
With these changes, I manage to get the core_visu_overpainting example with PyQt5 on Windows, python34. I get an error while rolling the mouse wheel:
```bash
Display3d class successfully initialized.
 ########################################
  File "core_visualization_overpaint_viewer.py", line 350, in wheelEvent
    if event.delta() > 0:
AttributeError: 'QWheelEvent' object has no attribute 'delta'
```

jf---:
@tpaviot , nice catch, thanks!

These suggested fixes have been committed...

There still are a few hairy things in the API...
For example:

```python
from OCC.Display.backend import get_backend, get_qt_modules, HAVE_PYQT5
print( HAVE_PYQT5) # None
```

That is, you have to import the backend before __HAVE_PYQT5__ is set...

```python
from OCC.Display.backend import get_backend, get_qt_modules, HAVE_PYQT5
backend = get_backend()
print( HAVE_PYQT5 ) # True
```

tpaviot:
@jf--- rather than a HAVE_PYQT5 global variable (it's not very elegant), a have_pyqt5 function, returning True or False, might be better.

jf---:
@tpaviot , thanks for reviewing... your comments are integrated...

jf---:
@tpaviot , anything holding back this PR?

tpaviot:
Trying to run hello_wolrd example:
```bash
l\pythonocc-core\examples>c:\python34\python core_helloworld.py
INFO:OCC.Display.backend:backend loaded: qt-pyqt5
INFO:OCC.Display.SimpleGui:GUI backend set to: qt-pyqt5
  File "core_helloworld.py", line 21, in <module>
  File "c:\python34\lib\site-packages\OCC\Display\SimpleGui.py", line 87, in ini
t_display
    from OCC.Display.qtDisplay import qtViewer3d
  File "c:\python34\lib\site-packages\OCC\Display\qtDisplay.py", line 29, in <mo
dule>
    from backend import get_qt_modules
ImportError: No module named 'backend'
```

tpaviot:
core_display_unittest fails:
```bash
D:\Devel\pythonocc-core\test>python core_display_unittest.py
['D:\\Devel\\pythonocc-core\\test', 'C:\\Python27\\lib\\site-packages\\pyemof-0.
11.0-py2.7.egg', 'C:\\Python27\\lib\\site-packages\\pyexcel-0.1.7-py2.7.egg', 'C
:\\Python27\\lib\\site-packages\\texttable-0.8.4-py2.7.egg', 'C:\\Python27\\lib\
\site-packages\\pyexcel_io-0.0.8-py2.7.egg', 'C:\\Python27\\lib\\site-packages\\
pyexcel_test_plugin-0.0.1-py2.7.egg', 'C:\\Python27\\lib\\site-packages\\pyexcel
_presentation-0.0.1-py2.7.egg', 'D:\\mysite-packages', 'D:\\Devel\\pythonocc-cor
e\\test', 'C:\\Windows\\SYSTEM32\\python27.zip', 'C:\\Python27\\DLLs', 'C:\\Pyth
on27\\lib', 'C:\\Python27\\lib\\plat-win', 'C:\\Python27\\lib\\lib-tk', 'C:\\Pyt
hon27', 'C:\\Python27\\lib\\site-packages', 'C:\\Python27\\lib\\site-packages\\F
ontTools', 'C:\\Python27\\lib\\site-packages\\PIL', 'C:\\Python27\\lib\\site-pac
kages\\itk', 'C:\\Python27\\lib\\site-packages\\win32', 'C:\\Python27\\lib\\site
-packages\\win32\\lib', 'C:\\Python27\\lib\\site-packages\\Pythonwin', 'C:\\Pyth
on27\\lib\\site-packages\\wx-2.8-msw-unicode']
pyside test
INFO:OCC.Display.backend:backend loaded: qt-pyside
INFO:OCC.Display.SimpleGui:GUI backend set to: qt-pyside
 ###### 3D rendering pipe initialisation #####
Display3d class initialization starting ...
Aspect_DisplayConnection created.
Graphic_Driver created.
WNT window created.
V3d_Viewer created.
V3d_View created
AIS_InteractiveContext created.
Display3d class successfully initialized.
 ########################################
pyqt4 test
CRITICAL:OCC.Display.backend:a backend is already loaded...``get_backend`` can o
nly be called once
  File "core_display_unittest.py", line 39, in <module>
    pyqt4_display, start_display, add_menu, add_function_to_menu = init_display(
'qt-pyqt4')
  File "C:\Python27\lib\site-packages\OCC\Display\SimpleGui.py", line 31, in ini
t_display
    USED_BACKEND = get_backend(backend_str)
  File "C:\Python27\lib\site-packages\OCC\Display\backend.py", line 70, in get_b
ackend
    raise ImportError(msg)
ImportError: a backend is already loaded...``get_backend`` can only be called on
ce
```

jf---:
Ah, so I added a check that no 2nd backend can be loaded, if one is already there...
-1 for API changes so let me fix

src/addons/Display/backend.py
```diff
@@ -0,0 +1,228 @@
...
+        msg = ("no Qt backend is loaded, hence cannot return any modules\n"
+               "either you havent got PyQt5, PyQt4 or PySide installed\n"
+               "or you havent yet loaded a backend with the "
+               "`OCC.Display.backend.get_backend` function")
+
+        raise ValueError(msg)
+
+
+def have_pyqt5():
+    global HAVE_PYQT5
```
tpaviot: shouldn't we make a check here, i.e. try to import Qt5 modules. Indeed if no display has been inited, these functions will always return False.

jf---:
So you're on py34 by default?
Nice...

> hello_world issue: 

probably because not using a full path eg, import backend, corrected: from OCC.Display import backend

> unittest

you cannot load all backend ( Qt backend are much mutex ) so refactored a test for each backend

tpaviot:
I reported an issue in on commit. I will deeply test this branch before merging, the Display part used to work quite well although it does not support PyQt5, I don't want any regression.

tpaviot:
I worked on your branch:
* merged all commits into one
* PEP8 fixes
* global variables removed everywhere t's not required
See branch https://github.com/tpaviot/pythonocc-core/tree/jf/tp/pyqt5
I suggest you close this one, and test jf/tp/pyqt5. a PR has been opened see #184

jf---:
Oufff, I did fix the wx thing but didnt push it...

PR #178, opened 2015-12-29
Conda support for travis

tpaviot:
This branch uses the conda dlr-sc repository to download/install all dependencies required by pythonocc. This makes the compilation much faster.

Changes also include a small fix for unittests on py3k. Everything should be ok now with travis.


jf---:
Very cool!

PR #182, opened 2016-01-06
Removed landscape checks

tpaviot:
* pythonocc-core is not really a python package. It's about 90% of C++ code. There's not so much python code to review

* pylint and pep8 checkers can  easily be ran in local mode

* landscape hosted solution does not work well.

* landscape flag is alqays brown or red. Boring.

For all those reasons, I suggest to remove landscape support.


jf---:
:+1:

PR #186, opened 2016-01-07
prepare release 0.16.3

tpaviot:
@jf---  @rainman110 get ready to upload your binaries to binstar as soon as this pr is merged. I'll announce the release on the ml once it's available using conda.

windows versions are being packaged.

after that, let'smove to oce-0.17.1 support.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -1,5 +1,7 @@
 pythonOCC is developed and maintained by Thomas Paviot (tpaviot@gmail.com).
 
 pythonOCC-core contains contributions from :
-* Jelle Feringa (jelleferinga@gmail.com) : conda build for OSX, examples, traits display
-* Guillaume Florent (florentsailing@gmail.com) : geometry airfoil example.
+* Jelle Feringa (jelleferinga@gmail.com): conda build for OSX, examples, PyQt5 support
+* Guillaume Florent (florentsailing@gmail.com): geometry airfoil example.
+* Martin Siggel (martin.siggel@dlr.de): conda recipes for linux/windows, wrapper fixes,
+matplotlib examples.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,7 +22,7 @@ set(PYTHONOCC_VERSION_MAJOR 0)
 set(PYTHONOCC_VERSION_MINOR 16)
 set(PYTHONOCC_VERSION_PATCH 3)
 #  Empty for official releases, set to -dev, -rc1, etc for development releases
-set(PYTHONOCC_VERSION_DEVEL -dev)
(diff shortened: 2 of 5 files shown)


jf---:
@tpaviot , I suggest a single point of download, which we point to on the conda badge, which is @rainman110 / DLR site.

Shall we keep doing so or make a separate "official" pythonocc conda account?
We can use that new account to collaborate on, since more users an share an account

tpaviot:
So far, it's ok for me to point to @rainman110 /DLR site. We'll see in the future, if ever DLR site is not maintained anymore. Unless @rainman110 wants to move to an official pythonocc repository, I prefer, for now, keep working this way.

jf---:
OK, I think that's the way to go, we'll speak about a pythonocc account for 0.17.1, if we feel there's a need for it at all ( I'm happy for now ) ?

rainman110:
@tpaviot, @jf--- I am currently building the release. Linux and Win are already done. Mac takes longer due to the rpath fixing.
It would be great, to have some automatic build process for the releases such that I don't have to do this manually on different machines. I think, we should also have an official pythonocc anaconda account for future releases. This way, you are independent from DLR specific patches and changes and also it just looks better :smile:

jf---:
@rainman110 good point...

TL;DR
* Use CI for CD ( cont. delivery ) -- lets not repeat the effort
* A commit to master is a new release

---

What about the following outlook:

* a commit to master triggers a travisci build, using conda
* whenever all tests pass, I think we should use __conda upload__ to push the resulting package to the anaconda site
* these builds should go to a separate channel, say, a dev channel
* whenever we want to release a specific package, just change the channel from dev -> main
* then run __conda update pythonocc-core --channel dev__
* this is why [automatic versioning](https://github.com/warner/python-versioneer) is important
  * in addition to Major, Minor, Patch, you'd have another attribute, SHA1, referring to the git commit
  * eg, 0.16.3.6f5280 / 6f52808fba5db219d57dfe8e739ded6e2653c2bc

now, currently, this would cover linux. we can have osx #187 for travis too. 
do the same thing with appveyor for windows

jf---:
@rainman110 , many thanks for the 0.16.3 builds 
awesome, thanks... :tropical_drink:

rainman110:
@jf--- Yes, this is absolutely the way to do it.

To sum up what remains to do:
 - OS X builds with travis
 - Make app-veyor functioning for the windows builds (last time i checked it was not working)
 - Use git SHA as version id. We need to add the SHA to the conda recipe and the CMakeLists.txt as well.

jf---:
@rainman110 Great... that we share the outlook...
let me move this conversation to a new issue, we've been discussing this idea a nr of times in various places...

PR #196, opened 2016-01-11
jf/pyqt5-fixes

jf---:
hotfix wheelEvent segfault on pyqt5 backend


jf---:
when running an example with pyqt5 that requires menus ( core_topology_glue.py for example ), it will segfaults on the old school signal / slots syntax.

testest on (osx) pyside, pyqt4, pyqt5

tpaviot:
tested

+1

PR #203, opened 2016-01-21
Enabled linux + osx travis matrix builds

rainman110:
This commit enables a travis matrix build job on OS X and Linux. All tests are passing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,10 @@
+sudo: false
+
 language: cpp
+
+os: 
+  - osx
+  - linux
 
 compiler:
   - gcc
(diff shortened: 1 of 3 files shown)


tpaviot:
Some more tests should be launched when the build is completed:
````bash
python core_webgl_unittest.py
xvfb-run -s "-screen 0 1024x768x16" python core_display_pyqt4_unittest.py
xvfb-run -s "-screen 0 1024x768x16" python core_display_pyqt5_unittest.py
xvfb-run -s "-screen 0 1024x768x16" python core_display_pyside_unittest.py
xvfb-run -s "-screen 0 1024x768x16" python core_display_wx_unittest.py
```
Indeed it's important to test the graphical part, and especially if the graphic driver is succesfully initialized. Don't know is xvfb is available on osx, maybe this should be run on linux only

tpaviot:
On travis, linux build is much faster than osx

tpaviot:
Ok, build passes. Let me merge the PR as is, we'll add missing tests in another PR. step by step.

rainman110:
> On travis, linux build is much faster than osx

I think the linux machines use docker and are more performant. The OS X ones are proper VMs and therefore slow.

PR #205, opened 2016-01-22
wx mouse wheel

tpaviot:
Patch from  from Emad Dlala adding mouse wheel zoom support.
7See this thread on pythonocc ml: https://groups.google.com/forum/#!topic/pythonocc/4Npv_ZGdlek


tpaviot:
Tested, works on Windows

PR #209, opened 2016-01-25
Added ifc2x3 example

tpaviot:
A simple IFC2x3 geomtry visualizer in 67 lines of code.

ifc2x3_example_screenshot


jf---:
This is __AWESOME__

But :-1: this requires a dependency ( the amazing IfcOpenShell ) and therefore should be added to pythonocc-contrib

tpaviot:
@jf--- The grid/boundingbox is a "graduated triehdron". It was added into occt650 afaik.

You're right this example requires ifcopenshell to be installed. I merged one of your recent contribution (core_visualization_graphic3d_custom_opengl.py) that requires numpy, because numpy is not necessary or even optional for pythonocc. What's the difference here with the ifc example ? Or do you think both these examples should be moved to pythonocc-contrib ?

jf---:
where does one draw the line? the fact that numpy is super well distributed doesnt change too much...

your point is correct: in __core_visualization_graphic3d_custom_opengl.py__ it would be best to offer a slow implementation when numpy is not available, where when numpy is not found, you just render say 10.000 lines, not a 1.000.000

with ifcopenshell, there is no possible backup... the dependency is a hard requirement...

Does this provide a reasonable suggestion? 

* [ ] Fix the numpy dependency in __core_visualization_graphic3d_custom_opengl.py__ ( that's on me )
* [ ] add the awesomely cool ifcopenshell example -> pythonocc-contrib...

---

and please dont get me wrong, I think that pythonocc + ifcopenshell is __precisely__ underscoring the relevance of OS sw dev :sunglasses:

jf---:
@tpaviot , thanks for the "graduated triehdron" hint, I recall using this sometime ago ;)

tpaviot:
@jf--- ok, I move this PR to contrib, please submit a PR for the core_visu example.

PR #210, opened 2016-01-27
DisplayVector method was broken due to a subtle missed API change

jf---:
DisplayVector method was broken due to a subtle missed API change

### DIFF
--- a/src/addons/Display/OCCViewer.py
+++ b/src/addons/Display/OCCViewer.py
@@ -247,8 +247,7 @@ class Viewer3d(OCC.Visualization.Display3d):
     def DisplayVector(self, vec, pnt, update=False):
         if self._inited:
             aPresentation = Prs3d_Presentation(self._struc_mgr)
-            arrow = Prs3d_Arrow()
-            arrow.Draw(
+            Prs3d_Arrow.Draw(
                 aPresentation.GetHandle(),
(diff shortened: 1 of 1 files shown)


tpaviot:
@jf---  Is there an example in our list where I could check the fix works ?

jf---:
Sure:

```python
from OCC.gp import *
display.DisplayVector(gp_Vec(100,0,0), gp_Pnt())
start_display()
```

tpaviot:
-1 gp_Vec should be imported at the beginning of occviewer

tpaviot:
@jf--- On my windows machine, only the cone is drawn, not the arrow body. Can you paste a screenshot of what you get on your mac ?

tpaviot:
@jf--- up

tpaviot:
@jf--- could you get the example work ?

jf---:
created another PR for this...
yep, example works, and alas, I checked the C++ code,  its just the Cone being drawn... too bad...

tpaviot:
@jf--what do you mean exactly with "yes the example works" and "only the cone being drawn". If only the cone is drawn, the example actually doesn't work imo

jf---:
so I read the C++ code and there is no body to render, its "just" the cone... ( -> its an ugly vector, but comes in handy... )

tpaviot:
ok, I got it

PR #211, opened 2016-01-27
Review/reg cb

tpaviot:
This PR adds a callback mechansim at the OCCViewer level. You can register as many callbacks as needed (callables) that will be called whenever a shape (or many shapes) are selected. This is completely GUI system indenpendant, and gives a basic way to interact with the display.

An example is provided that implements two callbacks, on of them computing the selected shape bounding box.


jf---:
:+1:

PR #213, opened 2016-01-27
numpy now optional for the custom opengl draw example

jf---:
numpy now optional for the custom opengl draw example


tpaviot:
@jf--- did you test the example with *and* without numpy ?

jf---:
sure

Issue #216, opened 2016-02-09
Three.js and X3Dom renderers produce invalid html on windows


rainman110:
The reason are invalid file urls to the javascript files. As a result, nothing is shown.

PR #217, opened 2016-02-09
Fixed invalid html in webgl files on windows

rainman110:
Fixed issue #216


tpaviot:
The fix from cfa4f88 supports python2 only, not py3k. Fix here 8070d95dd1b446c9d1b4f51ad14d46191da507bf

Issue #214, opened 2016-02-01
Discussion: Memory management


rainman110:
**This "issue" is aimed to be a discussion thread. I tried to write down my findings about potential problems of the current memory management. It should help me and others to understand, why the memory management is designed how it currently is.**

Memory management of the OCCT classes is somewhat difficult due to the ```Handle``` classes.

Handles take care of the life time of an object. If an object does not have any references to itself anymore (in terms of handles), it will be automatically deleted. Handles are basically smart pointers.

There are still memory issues left in pythonocc, most of them leaks.

Therefore I tried to understand, how the memory management is currently implemented in pythonocc and were are the problems originating from.

This brought some basic questions/issues:

  1. Why is the garbage collection required? There are other ways to deal with smart pointers and ownership of pointers than garbage collection.
  2. The garbage collection basically prevents the deletion of objects and shifts it to a later time. This brings up several problems, e.g. memory leaks if you forget to use the garbage collector.
  3. The way how the methods ```GetHandle``` are implemented looks strange. One particular issue with this approach is, that you could create a Handle of an object that lives on the stack. This only works in pythonocc currently, because you implicitly assume that swig always allocates objects on the heap. If a handle points on a stack object, it will try to delete its parent object, which will lead to a segfault. Why aren't handles create as in OpenCASCADE i.e. : 
 
        myHandle = Handle_Geom_BSplineSurface(Geom_BSplineSurface(...))

Because of these "issues", I played around with some toy project to better understand the memory management. It is completely possible to wrap the Handle* classes without any garbage collection and without any memory leaks.

When creating a handle to an object, the handle takes ownership of the lifetime of the object. At this point, also python/swig already has ownership of the object. Normally, this will lead to segfaults, as both python/swig and the handle try to delete the object. By using 

     MyObject.thisown = False

it is possible to remove the ownership from python/swig. In swig, the change of an ownership can be even defined in the interface file using code like:

```c++
%apply SWIGTYPE *DISOWN { A *adisown };
    
class Handle_A
{
public:
  Handle_A(A* adisown) {
   a = adisown;
   if (a) a->IncRefs();
  }

private:
A* a;
}
```
Unfortunetly, this feature is hardly documented in swig.

tpaviot:
@rainman110 Thanks for opening this thread. I spent hours working about memory management issues in pythonocc.

[...]

PR #220, opened 2016-02-11
Memory management improvements

tpaviot:
This PR changes the way pythonocc objects are freed/deallocated. It makes the GarbageCollector outdated and unnecessary, and a better memory usage according to valgring reports.

See issue #214 for a detailed explanation about this implementation.


rainman110:
@tpaviot Do you understand why the gui tests are failing sometime?

I don't think  that this has to do with the new memory management.

tpaviot:
@rainman110 It has nothing to do with memory management, but rather with the way gui tests are launched. On travis I use xvfb to run the test, from the command line, but it seems a more robust solution should be used, see https://docs.travis-ci.com/user/gui-and-headless-browsers/. I'm not surprised that this problem occurs, but I didnt have time to modify the xvfb code yet.

tpaviot:
@rainman110 According to travis doc, ```Note: Don’t run xvfb directly, as it does not handle multiple concurrent instances that way.``` Currently, xvfb is actually run directly.

tpaviot:
@rainman110 Trying to fix travis GUI issues. See #221. I'll rebase this one onto master as soon as the fix is merged.

rainman110:
It seems, that the build failures have indeed something to do with the new memory management. It's always the python 3.5 which makes problems only with the new memory management.

We should probably have a look to this particular configuration, i.e. python 3.5 on linux.

tpaviot:
Yes, the issue definitely comes from the memory management changes. Strange that only linux/py35 is affected. I have no idea/clue about what's going on.

rainman110:
@tpaviot I can reproduce the problems on my local machine. Strangely, there is actually no problem except from this error message. 

To provoke this error, it suffices to run the init_display() function from SimpleGui. That is, we don't need to add a shape, just the initialization suffices. In  the SimpleGui, there are not many OCC specific imports.

I don't have yet any clue, what is going on, but I'll figure it out.

tpaviot:
@rainman110 This maybe either to handles in Visualiation.h and OCCViewer.py::

* in Visualisation.h (see https://github.com/tpaviot/pythonocc-core/blob/master/src/Visualization/Visualization.h#L47), handles to V3d_Viewer and V3d_View are defined ;

* in OCCViewer.py, I had to store handles in order to prevent segfaults related to objects destruction. See https://github.com/tpaviot/pythonocc-core/blob/master/src/addons/Display/OCCViewer.py#L136

rainman110:
@tpaviot Thanks for the hints!

What might be again problematic here is the return by reference of the handles in Visualization.h. I'll try to change this and let's see what it helps.

Yesterday, I made a valgrind analysis where I could not find anything different between the old and the new memory management. The Problem lies probably in the handling of the OpenGL context.

I suppose, that when python ends, it calls the desctructors of all remaining objects, which might kill the view too early and the scene still tries to render something. This would explain the bad_drawable errors.

Unfortunately, I don't have much time currently to dig into it deeper. I'll let you know as soon I have found the root of the problem.

rainman110:
Hi @tpaviot It's been a while ago. Today I had the chance to look into this issue. I could not properly understand, why this issue happens, but I found a workaround to prevent it.

Just add the line
```python
self._display.thisown
```

into the method ```qtViewer3d.InitDriver```.

I suspect the issue is, that there is a top level Qt OpenGl widget, which would get destroyed after destroying the V3d_View. This means, that the V3d_View might destroy the OpenGl context of the OpenGl Widget, which itself tries to draw something. This then causes the X error.

By adding the mentioned line, we delay/hinder the destruction of the Visualization class. Of course, this creates a small memory leak, which should be negligible as it is executed only once.

In order to improve the situation, one had to rewrite the Qt Viewer to not use an OpenGl widget (as done in the c++ qt example of opencascade).

tpaviot:
@rainman110 Great! That worked. I spent most of the past couple of weeks trying to figure out what's happening.

The current PR can be merged.

jf---:
Congrats @rainman110 & @tpaviot , this is a big deal and I can only imagine pretty tough work to plough through... great to see this come about...

PR #234, opened 2016-04-05
Review/prepare 0.16.4

tpaviot:
After this branch is merged, release 0.16.4 and the new memory management system


jf---:
Think that's a wise move to make, its a pretty significant chance indeed...

PR #237, opened 2016-04-06
_XCAFPrs.so & 	_STEPCAFControl.so would not compile due to missing package: TKV3d

jf---:
_XCAFPrs.so & 	_STEPCAFControl.so would not compile due to missing package: TKV3d


tpaviot:
@jf--- Shouldn't this commit be merged into the vtk branch ?

jf---:
@tpaviot , uhm, nope, this is wrt OCE 0.17 / these dependencies don't relate to VTK

tpaviot:
I thought is was related to the recent vtk changes. You mean the temp-0.17 branch does not compile on your machine without this fix ?

jf---:
correct...

with ["my travisci"](https://travis-ci.org/jf---/pythonocc-core) all of the build matrix passes for this branch...
not sure why there's a difference between our respective builds...

tpaviot:
Strange indeed, maybe a difference with the compiler/linker versions ... Anyway, this fix is a really small change, if it does the job, ok for merging

jf---:
@tpaviot , so travisci uses Clang 6.x.x, where a XCode is Clang 7.x.x 
Clang is becoming more strict with newer releases, so quite likely that this is the cause.
thanks for merging.

PR #243, opened 2016-04-08
ported topology examples

jf---:
somehow really cool bool ops example were missing ( emmenthaler still is... )


tpaviot:
Thank you for this contribution. Examples have to be ported to recent changes to master, they have to be fixed, they have to be refactored and splitted into atomic ones, glue solid and glue solid edges are already available in the example suite. I took the branch and work on it https://github.com/tpaviot/pythonocc-core/tree/jf/boolops_example

jf---:
hi @tpaviot , OK, I'm happy to take care of the suggested corrections as well...

PR #248, opened 2016-04-10
jf/shader_

jf---:
This is work-in-progres, meant to discussing progress regarding this issue

---

close to having a working shader example
issue seems that `Graphic3d_ShaderProgram_Handle` ( which is what `aspect.ShaderProgram()` returns ) is not wrapped, therefore the shader program might not be set.

I'm uncertain of this, since `aspect.SetShaderProgram(aspect.ShaderProgram())` completes, while the following completes ( with no difference in visual result )

```python
type(aspect.ShaderProgram())
SwigPyObject
```

So the output is a correct type, but probably not doing anything?

OCE Shader location: /Users/jelleferinga/miniconda/envs/PyRAPID_QML/share/oce-0.17/src/Shaders
swig/python detected a memory leak of type 'Graphic3d_ShaderObject_Handle *', no destructor found.
swig/python detected a memory leak of type 'Graphic3d_ShaderObject_Handle *', no destructor found.
 Graphic3d_ShaderProgram_Handle is not wrapped, returned instance instead is a` <type 'SwigPyObject'>`


jf---:
hi @tpaviot ,

Would you be able to take a glance at this PR when 0.16.5 is released?
I think it might be a relatively minor issue with a class not being exposed in SWIG that stops us from unlocking GLSL shaders...

tpaviot:
There are two many commits here, please first rebase on top of the base branch

PR #255, opened 2016-04-23
added core_topology_traverse example

tpaviot:
This commit adds the Topo class as an example about how to explore and traverse topology


jf---:
hi @tpaviot , 

this looks cool -- what might be an interesting augment is that for instance show the 3 edges incident to a vertex ( just one out of many examples, show the 2 faces sharing an edge ), this would be a more intuitive approach, which could be a cool addition.

PR #256, opened 2016-04-24
qt backend fix

### LINKED ISSUE #224: Display get_backend Import error in Spyder
Hi,

I have been using PythonOCC from the Spyder IDE for a while, and have always closed the Qt window and rerun the script and restarted the display when I want to remake the geometry - However with some recent changes, the get_backend function in /Display/backend.py throws an error on the second run as the IPython kernel has already called get_backend, forcing me to restart the kernel. I initialize the display with:

start_display()

Should I be changing my workflow and not calling the usual display initalizer on each run, or is this an IPython/IDE-related bug? 

Thanks

tpaviot:
* the ```get_backend```function name is replaced with ```load_backend```, which is more explicit

* if the ```load_backend``` function is called more than once, the second times it returns an info message and the string of the load backend.

The behaviour is now:

```
>>> load_backend("qt-pyqt4")
INFO:OCC.Display.backend:backend loaded: qt-pyqt4
'qt-pyqt4'
>>> load_backend("qt-pyqt5")
INFO:OCC.Display.backend:The qt-pyqt4 backend is already loaded...``load_backend
y be called once per session
'qt-pyqt4'
>>>
```

Should fix #224

### DIFF
--- a/src/addons/Display/SimpleGui.py
+++ b/src/addons/Display/SimpleGui.py
@@ -21,7 +21,7 @@ import logging
 import sys
 
 from OCC import VERSION
-from OCC.Display.backend import get_backend, get_qt_modules
+from OCC.Display.backend import load_backend, get_qt_modules
 
 log = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


tpaviot:
Fix #224

jf---:
👍

Issue #244, opened 2016-04-08
automatic test of examples


jf---:
we've been thinking about using the examples as unit tests and to be able to run these automatically.

- to create test from the examples is easy, just add assert statement
- to bypass running the GUI, I suggest using a decorator for the functions to be tested
  - the decorator sets os.setenv("TEST_PYOCC", 1)
  - now, when loading `from OCC.Display.SimpleGui import init_display` we just return a mock object instead of a OCC.Display instance. Any method called simple wont perform any actions

test coverage of our examples would be a great thing to have, especially when working towards a new OCE release.

mo' code coverage mo' better...

tpaviot:
test at https://github.com/tpaviot/pythonocc-core/tree/review/run-examples-as-tests

jf---:
Cool! 
Happy to hear you think its worthwhile to further explore...
I'll start refactoring & placing __assert__ statement ( "assert SomeOperation.IsDone() ) for example

tpaviot:
Done. See #258

PR #258, opened 2016-04-26
Run examples as tests

tpaviot:
This branch adds a feature to run the example suite as tests.

Every single example in the src/example is ran in a "blind gui" mode, allowing examples to be executed from a console without any gui needed.

Those tests are launched after the unittest suite passes.

The tests have been fixed on both py2 and py3 so that travis builds fine.

Checked on win32/mingw32.

Fix isse #244


jf---:
Awesome, I think this will be quite an advantage in our effort working towards 0.17...

Issue #259, opened 2016-04-28
Standard errors like Standard_OutOfRange not caught


jf---:
For example:

```python
from OCC.gp import *
d = gp_Dir()
d.Coord(-1) # 1,2 or 3 are valid indices
# KABOOM
libc++abi.dylib: terminating with uncaught exception of type Standard_OutOfRange
```

Could this possible be compiler related?

tpaviot:
Same issue on windows/mingw, it's not compiler specific

tpaviot:
It used to work, I can't understant when it started to fail, and why. A quick check doesnt't show any obvious clue. I don't know if something changed in occ exception handling

jf---:
hi @tpaviot , yes, I recall on earlier version that pyocc wrapped these as RuntimeErrors...
So possible, this is a meaningful future unittest?

this comes up at the [forum](http://www.opencascade.com/content/how-catch-exceptions):

> You might want to check your compile flags. For some operations, CASCADE does not really throw a proper exception and performs as you describe. You must compile without the -DNO_CXX_EXCEPTION flag. If you have that flag set, CASCADE will abort instead of throwing the exception. This took us several days to find.

tpaviot:
Seems related to #13.

Everything is compiled with the right flag OCC_CONVERT_SIGNALS, the try/catch is done as the official doc stands. I suspect a conflict with builtin siwg exception, I'll check that

jf---:
Shall we close either issue, and mark it duplicate?
Right, wasnt this around the time that we've moved from SWIG 2 -> 3?

tpaviot:
* this is not related to the SWIG version (I tried with both 2 and 3, same issue)

* it's not related to the builtin SWIG exception handling system (tested with and without, same issue).

jf---:
Thanks for the heads up... that's quite a nr of possibilities you've excluded...
Hmm, so your suspicion is directed towards OCE?

jf---:
the other day @rainman110 offered to fire up valgrind for the failing #124 ( osx specific, thanks btw! ). possible this issue might be of greater urgence?

tpaviot:
I tried to compile in RelWithDebInfo mode, rather than MinSizeRel, no change.

Not sure about the issue priorites, it depends on so many factors to be fully objective, let's talk about that in another issue if necessary.

rainman110:
@tpaviot This error is totally valid. C++ throws an exception by accessing an invalid array index.

[...]

PR #262, opened 2016-05-01
Fixed issue #259

rainman110:
Reason for the segfault was calling strlen on a null pointer string. I changed the code to be more robust using std::strings.

Fixes issue #259


tpaviot:
Perfect, that works. I was far far away to imagine the issue comes from the string code. Thanks !

jf---:
Wow, that's fantastic that this critical issue has been solved in _such_ short notice!
Gents, should we include a unittests such to catch regressions? ( the gp_Dir().Coord(-1) is perhaps reasonable succinct )

rainman110:
Yes, a small test case would be definetely a good idea. :+1:

tpaviot:
Yes @jf--- a unittest is necessary, thanks for taking care of that

jf---:
Coming up...

PR #270, opened 2016-05-17
Review/prepare release 0.16.5

tpaviot:


jf---:
hi Thomas, 

Thrilled that we have a monthly release schedule once again, that's cool...
How should we go about continuing on the 0.17 release?
Perhaps a rebase of the branch is called for with the 0.16.5 release / exceptions fix?

PR #276, opened 2016-05-28
move SwapBuffers()

johanesmikhael:
move  SwapBuffers() to prevent multiplying drawn rectangle

### DIFF
--- a/src/addons/Display/qtDisplay.py
+++ b/src/addons/Display/qtDisplay.py
@@ -155,13 +155,13 @@ class qtViewer3d(qtBaseViewer):
     def paintEvent(self, event):
         if self._inited:
             self._display.Context.UpdateCurrentViewer()
-            # important to allow overpainting of the OCC OpenGL context in Qt
-            self.swapBuffers()
 
         if self._drawbox:
             self.makeCurrent()
             painter = QtGui.QPainter(self)
             painter.setPen(QtGui.QPen(QtGui.QColor(0, 0, 0), 1))
+            # important to allow overpainting of the OCC OpenGL context in Qt
+            self.swapBuffers()
             rect = QtCore.QRect(*self._drawbox)
             painter.drawRect(rect)
             painter.end()


jf---:
hi @johanesmikhael , IIRC buffers should be swapped only once per view update

also, the 2nd call you've added, shouldnt this take place after the drawing has completed, eg as the last statement?

something like:

```python

     def paintEvent(self, event):
          if self._inited:
              self._display.Context.UpdateCurrentViewer()
  
          if self._drawbox:
              self.makeCurrent()
              painter = QtGui.QPainter(self)
              painter.setPen(QtGui.QPen(QtGui.QColor(0, 0, 0), 1))
              rect = QtCore.QRect(*self._drawbox)
              painter.drawRect(rect)
              painter.end()

         self.swapBuffers()
```

that said, overpainting the rectangle works perfectly on osx. on which platform are you working?

johanesmikhael:
hi @jf--- , the i've been tested your solution but still doesn't give desirable "selection rectangular" on my screen. I actually look up on example/core_visualization_overpaint_viewer.py on its _overpaint() function an see swapBuffers() placed just before QPainter draw method called. Here is the code that i mentioned: 
`def paintEvent(self, event):
        """ handles all actions that redraw the viewport

        Parameters
        ----------
        event : QPaintEvent

        See Also
        --------
        this method is also invoked through the self.update method, see the
        mouseMoveEvent method
        """

        if self._inited:
            # actions like panning, zooming and rotating the view invoke
            # redrawing it
            # therefore, these actions need to be performed in the paintEvent
            # method
            # this way, redrawing the view takes place synchronous with the
            # overpaint action
            # not respecting this order would lead to a jittering view
            if not self._dispatch_camera_command_actions():
                # if no camera actions took place, invoke a redraw of
                # the viewport
                self._display.View.Redraw()

            if self.context().isValid():
                # acquire the OpenGL context
                self.makeCurrent()
                painter = QtGui.QPainter(self)
                painter.setRenderHint(QtGui.QPainter.Antialiasing, True)
                # swap the buffer before overpainting it
                self.swapBuffers()
                # perform overpainting
                self._overpaint(event, painter)
                painter.end()
                # hand over the OpenGL context
                self.doneCurrent()
            else:
                print('invalid OpenGL context: Qt cannot overpaint viewer')`

and its worked to eliminate multiple rectangle that drawn on screen. I working on Windows anyway

johanesmikhael:
Hi @jf--- I tried your codes and still results multiple rectangular drawn on my screen. I actually modify the code based on this example on example/core_visualization_overpaint_viewer.py paintEvent() method:

``` python

    def paintEvent(self, event):
        if self._inited:
            # actions like panning, zooming and rotating the view invoke
            # redrawing it
            # therefore, these actions need to be performed in the paintEvent
            # method
            # this way, redrawing the view takes place synchronous with the
            # overpaint action
            # not respecting this order would lead to a jittering view
            if not self._dispatch_camera_command_actions():
                # if no camera actions took place, invoke a redraw of
                # the viewport
                self._display.View.Redraw()

            if self.context().isValid():
                # acquire the OpenGL context
                self.makeCurrent()
                painter = QtGui.QPainter(self)
                painter.setRenderHint(QtGui.QPainter.Antialiasing, True)
                # swap the buffer before overpainting it
                self.swapBuffers()
                # perform overpainting
                self._overpaint(event, painter)
                painter.end()
                # hand over the OpenGL context
                self.doneCurrent()
            else:
                print('invalid OpenGL context: Qt cannot overpaint `viewer')```

The SwapBuffers() is placed just before QPainter draw method called, and its working on me to create clean "selection rectangular" on my qtDisplay screen. I working on Windows 64 bit anyway,, using Python 2.7 32 bit.

tpaviot:
@johanesmikhael On Windows 32/py27 I can't see any difference between all the modifications you suggest and the current core_visualization_overpaint example. This is certainly specific to an issue with your OpenGl driver. What's your graphic device ?

johanesmikhael:
I guess is not the driver. I realize the swapBuffers() should be called inside makeCurrent() and doneCurrent() before QPainter draw called. In my case here is the difference:

Before modification:

rectangle1

After modification: 
rectangle

sorry for the confusion, I realize it when try to hook pythonocc display with ifcopenshell, I have to check if ifcopenshell creates the problem instead.

tpaviot:
@johanesmikhael I close the issue, please feel free to reopen is ever the problem remains

PR #304, opened 2016-09-15
CI/CD changes

tpaviot:


jf---:
hi @tpaviot , so is the renewed Appveyor a better experience to work with?
I'm seeing you work very hard on a lot of CI/CD related aspects...
Is it getting closer to the polished experience of Travis?

tpaviot:
@jf--- It's much better, yes, but not perfect unfortunately. I prefer merging this branch as is, rather than delaying the 0.17 release. Appveyor support is not that import for end users, just for us.

PR #316, opened 2016-09-26
jf/signals_slots_example

jf---:
signal/slot example

### DIFF
--- a/examples/core_display_signal_slots.py
+++ b/examples/core_display_signal_slots.py
@@ -2,11 +2,13 @@ from __future__ import print_function
 import sys
 
 from OCC.BRepGProp import brepgprop_LinearProperties
-from OCC.BRepPrimAPI import BRepPrimAPI_MakeBox
+from OCC.BRepPrimAPI import BRepPrimAPI_MakeBox, BRepPrimAPI_MakeSphere
 from OCC.Display.SimpleGui import init_display
 from OCC.Display.backend import get_qt_modules
 from OCC.GProp import GProp_GProps
 from OCC.TopAbs import TopAbs_SOLID, TopAbs_EDGE, TopAbs_FACE
+from OCC.TopLoc import TopLoc_Location
+from OCC.gp import gp_Trsf, gp_Vec
 
 display, start_display, add_menu, add_function_to_menu = init_display("qt-pyqt5")
 QtCore, QtGui, QtWidgets, QtOpenGL = get_qt_modules()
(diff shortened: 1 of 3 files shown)


tpaviot:
@jf--- please check travis log to see what's wrong

jf---:
```bash
  File "../examples/core_display_signal_slots.py", line 12, in <module>
    QtCore, QtGui, QtWidgets, QtOpenGL = get_qt_modules()
  File "/Users/travis/miniconda/envs/_test/lib/python2.7/site-packages/OCC/Display/backend.py", line 238, in get_qt_modules
    raise ValueError("no backend has been imported yet with "
ValueError: no backend has been imported yet with ``load_backend``... 
[failed]
```
OK, need to be more careful with loading the GUI...

jf---:
@tpaviot passing CI ;)

tpaviot:
@jf--- I'm not convinced by this fix. There should not be any Travis specific behaviour, otherwise we can't predict how it will run on our machines. The test has to be run
Is this example qt5 specific ?

jf---:
@tpaviot , yes its Qt specific...
so i understand that when running the test, a 'null' backend is loaded, sidestepping any gui side-effects? moreover, this example requires you to click... i think integrating CI here is a little academic?

tpaviot:
@jf--- My question was is it qt *5* specific ? i.e does it work only with qt5 or for both qt4/qt5 ? A clean CI is a major achievement of the past months, I really want everything to be tested on travis

jf---:
@tpaviot , no doubt the CI work is a *huge* deal 🏆 
it could work on both qt4 and qt5, yep...

i'm really not very happy with our scruffy qt ( 4, 5, pyside ) abstraction... its hard to test ( requires 3 platforms to checkout )

let's use [qtpy](https://github.com/spyder-ide/qtpy) which is really the right way of doing so...
the > 250 commits tell us its not easy to get right ;)

are you OK if i start abstracting Qt like this? really dislike our own solution...

tpaviot:
Adding another dependency is not a priority. So far, the qt abstraction just works, to me it's ok. There are many other WIP (I think about the GLSL work, the QML example you promised a long time ago ;-) ).

jf---:
OK, on it ;)

Yes, the QML stuff is really exciting 🚀 

image

Imagine how cool it could look with GLSL shader 🎱

tpaviot:
@jf---  Maybe soon a blog post about this work ?

jf---:
@tpaviot , yes, this would be fun to write about... 
it'll still take me a few weeks to complete it...
thanks for the invite...

tpaviot:
@jf--- Can you please move the line `from OCC.Display.backend import get_qt_modules` *after* the display is initialized. I think this will make travis pass, no need to bypass it.

jf---:
OK, 1st thing tomorrow, thanks for the heads up!

tpaviot:
@jf--- No, after all, you should tell pythonocc to remove this test from the list of tests to exclude from run_example_as_text. For that, you just have to add core_display_signal_slots.py to https://github.com/tpaviot/pythonocc-core/blob/master/test/run_examples_as_tests.py#L30

After that, you can rebase -i onto master to remove commit fd83726

I'll merge as soon as it is done

jf---:
coming up, thanks for the detailed explanation...l

jf---:
`The job exceeded the maximum time limit for jobs, and has been terminated`, 1/6 fine otherwise

tpaviot:
@jf--- It uses to occur on travis, I generally have to manually trigger another build.

Issue #329, opened 2016-09-28
enable rendering of contours on top of shaded geometry


jf---:
the edges make the topology a lot more understandable

thanks @lalvarezt for pointing this out...

![image](https://cloud.githubusercontent.com/assets/620988/18924361/38761ce4-85b0-11e6-8340-196587ee5650.png)

tpaviot:
@jf--- can u submit a PR ?

tpaviot:
Fixed by PR #330

PR #330, opened 2016-09-28
improved default tesselation, face boundaries drawn by default, selection orange by default, phong shading by default.

### LINKED ISSUE #328: enable phong shading by default
smoother hilights, performs interpolation of the normals where the vertices meet

image

jf---:
improved default tesselation, face boundaries drawn by default, selection orange by default, phong shading by default.

fixed #328 #329

### DIFF
--- a/src/addons/Display/OCCViewer.py
+++ b/src/addons/Display/OCCViewer.py
@@ -131,7 +131,7 @@ class Viewer3d(OCC.Visualization.Display3d):
         self.View.ZFitAll()
         self.View.FitAll()
 
-    def Create(self, create_default_lights=True):
+    def Create(self, create_default_lights=True, draw_face_boundaries=True, phong_shading=True):
         self.Init(self._window_handle)
         self.Context_handle = self.GetContext()
         self.Viewer_handle = self.GetViewer()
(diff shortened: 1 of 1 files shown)


tpaviot:
@jf--- I disagree with selection orange by default. I can't see anything selected, it's blended with shape color. This is not related with #328 and #329.

tpaviot:
see attached screenshot why orange selection color is bad
capture
The draw option boundary is quite cool !

jf---:
OK, I'll comment out the selection color, rather than delete it... happy to hear you like the boundary option too...

tpaviot:
@jf--- The selection color orange is already commented out in OCCViewer.py, not need to do it twice (see https://github.com/tpaviot/pythonocc-core/blob/master/src/addons/Display/OCCViewer.py#L148).

jf---:
Yes, so reverted to that state in this commit

tpaviot:
ok I see

wrt the selection color should be moved to another issue ticket "default colors", that would change default background color, default shape color as well as default selection color. We must ensure all of them produce a nice rendering. I agree that current default colors are quite old and that new colors would be welcome.

Issue #324, opened 2016-09-27
TBB related issue building current master on osx


jf---:
@tpaviot , congrats on the recent refactor of the CI / conda scripts, its great to be able to use these locally, really neat...

that said, while building I've ran into the following; it seems that intel TBB is linked.
this is peculiar:

- it seems like there are no flags to set / remove linking to TBB
- TBB is not part of the dependencies in `meta.yaml`
- FYI I'm using DLR-SC's OCE 0.17.2 build, which utilizes TBB
- see the img at the bottom, wrt how TBB's path

```python
+ python run_tests.py
Traceback (most recent call last):
  File "run_tests.py", line 22, in <module>
    import core_wrapper_features_unittest
  File "/Users/jelleferinga/miniconda/conda-bld/work/test/core_wrapper_features_unittest.py", line 25, in <module>
    from OCC.Standard import Standard_Transient, Handle_Standard_Transient
  File "/Users/jelleferinga/miniconda/envs/_test/lib/python3.5/site-packages/OCC/Standard.py", line 22, in <module>
    _Standard = swig_import_helper()
  File "/Users/jelleferinga/miniconda/envs/_test/lib/python3.5/site-packages/OCC/Standard.py", line 21, in swig_import_helper
    return importlib.import_module(mname)
  File "/Users/jelleferinga/miniconda/envs/_test/lib/python3.5/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
ImportError: dlopen(/Users/jelleferinga/miniconda/envs/_test/lib/python3.5/site-packages/OCC/_Standard.so, 2): Library not loaded: libtbbmalloc.dylib
  Referenced from: /Users/jelleferinga/miniconda/envs/_test/lib/python3.5/site-packages/OCC/_Standard.so
  Reason: image not found
TESTS FAILED: pythonocc-core-0.17-py35_5
```

![image](https://cloud.githubusercontent.com/assets/620988/18870114/8b6c39c6-84af-11e6-9541-002dd813d2ea.png)

![image](https://cloud.githubusercontent.com/assets/620988/18870323/5eaf0bec-84b0-11e6-90fc-3cdd34c1f2a5.png)

tpaviot:
@jf--- Intel TBB is not used directly by pythonocc, no pythonocc library should be linked to TBB. Maybe this has something to do with @rainman110 compilation for osx64.

rainman110:
[...]

PR #331, opened 2016-09-29
Install swig modules properly

rainman110:
By installing the modules instead of copying them, cmake can do its magic and set the rpaths correctly.

This should fix issue https://github.com/tpaviot/pythonocc-core/issues/324. 

Note: I am currently assuming, that the library dependencies are all installed in ```${CMAKE_INSTALL_PREFIX}/lib```. This can be different on rpm based systems (i.e. lib64). It would be good, if we could additionally provide a cmake configuration value, where we can set/override the library path manually.


tpaviot:
@rainman110 Does it change anything to the cmake min required version ?

rainman110:
@tpaviot Not that I know. Setting the variables should be no problem, even if cmake does not know them. The install(TARGETS... ) thing should also be working.

But I haven't tried to use it with an old (2.6) cmake I have to admit. Would it harm, to force a cmake 2.8.12 or something? This is already pretty old I think.

tpaviot:
Currently, the minimum cmake required version is 2.6 but we can move up to 2.8 of course

jf---:
+1 for using recent CMake versions, 3.5 is the current one

Issue #341, opened 2016-10-17
Default SimpleGui uses a quarter of the space in the application's window


henrybaxter:
https://www.dropbox.com/s/f3jrbz2mze0hr8y/Screenshot%202016-10-16%2021.44.40.png?dl=0

This is on a retina display, and as I write this, I wonder if that's why? Anyway the following code gives the screenshot above:

```
from OCC.Display.SimpleGui import init_display
from OCC.BRepPrimAPI import BRepPrimAPI_MakeBox

display, start_display, add_menu, add_function_to_menu = init_display()
my_box = BRepPrimAPI_MakeBox(10., 20., 30.).Shape()

display.DisplayShape(my_box, update=True)
start_display()
```

That is, the viewer only uses the bottom left quarter of the window.

Note the code above is copied from the "hello dumb box" tutorial.

I would be more than happy to provide more information or try to fix it if someone has an idea.

henrybaxter:
Also note that it scales with the resize value, so eg
    ... = init_display('qt-pyqt5', (200,200))
gives a smaller window, and still the viewer uses the lower left quarter.

henrybaxter:
Perhaps this bug (and fix) is related, using Graphic3d_RenderingParams::Resolution to 'display correctly fixed size elements on hidpi screens' https://tracker.dev.opencascade.org/view.php?id=25350 ?

henrybaxter:
I am unable to access any member of the V3d_View's Graphic3d_RenderingParams, eg

        p = self.View.RenderingParams()
        print(p.Resolution())
    AttributeError: 'Graphic3d_RenderingParams' object has no attribute 'Resolution'

Not sure what this would tell me if I could. I do know that the View_3d's Size() and the QMainWindow frameSize() both return the logical pixel size of the main window on the screen. I feel like the View_3d somehow has to find out it should render 2x bigger. Is that the View_3d's job, or some other object?

jf---:
hi @henrybaxter , so that fix you mention if for OCCT 7.0.1, while the version exposed in pythonocc corresponds to 6.8.0

I've dabbled with this problem myself.
I think the easier solution, well workaround might be to get the NSView from Qt, and set the correct arguments from that object.

`Cocoa_Window` [is constructed here](https://github.com/tpaviot/pythonocc-core/blob/master/src/Visualization/Display3d.cpp), so an alternative might be call the methods for hi-dpi there?

Hope that helps to get some grip on the issue...

[...]

PR #359, opened 2016-11-17
Fix of the resizing issue of the qtViewer widget

adam-urbanczyk:
As mentioned on the google group this fixes the issue for ugly resizing of the viewer. 

It is tested on PyQt4.

python_occ_resizing_issue

### DIFF
--- a/src/Display/qtDisplay.py
+++ b/src/Display/qtDisplay.py
@@ -90,6 +90,7 @@ class qtBaseViewer(QtOpenGL.QGLWidget):
         return win_id
 
     def resizeEvent(self, event):
+        super(qtBaseViewer,self).resizeEvent(event)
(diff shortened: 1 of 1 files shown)


tpaviot:
@adam-urbanczyk is it osx specific ? is it related to #341 ?

adam-urbanczyk:
I experienced the issue on Linux. If this is also a problem on OSX I cannot check. It definitely might be - but that depends on details of (Py)Qt implementation for OSX.

I think calling the superclass method for resizing first is generally a good idea (i.e. definitely should not hurt), but would be great if someone could test on different platforms.

jf---:
Gents, not this is not related to #341, which has to do with pixel density.
This is a fix to a missing call to the super class.
👍

PR #367, opened 2016-12-16
Update README.md

jf---:
Currently, the conda install command on the README.md has two issues:

1. by default legacy python is installed ( 2.7, rather than 3.5 )
2. it would make sense to create an environment rather than install in conda's root environment, which is best practice


tpaviot:
ok

PR #368, opened 2016-12-16
WIP; GLSL example requires pythonocc-core>=0.17.1

jf---:
minimal, but working GLSL example
requires pythonocc-core>=0.17.1


jf---:
this is a WIP example, to revamp the GLSL discussion...
this is the simplest possible working example, which doesnt require changes to the `Visualization` module, nor depends on shader files...

jf---:
[this is quite interesting](http://www.cppblog.com/eryar/archive/2015/07/21/211331.html), where Qt's API is used to set the shader, rather than OCE's...

tpaviot:
@jf--- you can read chinese ? quite impressive  😸

jf---:
Well, google translate integration makes being a global citizen straightfwd 😉

tpaviot:
According to me this example is not minimal. It becomes minimal in the following case:
```
from OCC.AIS import AIS_Shape
from OCC.BRepPrimAPI import BRepPrimAPI_MakeSphere
from OCC.Graphic3d import (Graphic3d_ShaderProgram, Graphic3d_TOS_VERTEX, Graphic3d_TOS_FRAGMENT,
                           Graphic3d_ShaderObject)
from OCC.TCollection import TCollection_AsciiString

shape = BRepPrimAPI_MakeSphere(100).Shape()
anIO = AIS_Shape(shape)
display.Context.Display(anIO.GetHandle())

# gragment shader
fs = """
void main(void)
{
    gl_FragColor=vec4(0.0, 1.0, 0, 1.0);
}
"""

# vertex shader
vs = """
void main()
{
    gl_Position = occProjectionMatrix * occWorldViewMatrix * occModelWorldMatrix * occVertex;
}
"""

# construct the shader, load, compile and attach the GLSL programs
vs_shader = Graphic3d_ShaderObject.CreateFromSource(Graphic3d_TOS_VERTEX, TCollection_AsciiString(vs))
fs_shader = Graphic3d_ShaderObject.CreateFromSource(Graphic3d_TOS_FRAGMENT, TCollection_AsciiString(fs))
aProgram = Graphic3d_ShaderProgram()
aProgram.AttachShader(fs_shader)
aProgram.AttachShader(vs_shader)

# attach the shader to the AIS_Shape representation that renders the sphere
aspect = anIO.Attributes().GetObject().ShadingAspect().GetObject().aspect().GetObject()

if aProgram.IsDone():
    aspect.SetShaderProgram(aProgram.GetHandle())
else:
    raise AssertionError("no valid shader program found")

# when a shader fails to compile, raise an AssertionError
assert not aspect.ShaderProgram().IsNull(), "no shader program is null"

# redisplay the sphere, when the shader was attached to its AIS_Shape aspect
display.Context.Redisplay(anIO.GetHandle())
display.FitAll()
start_display()
```
This example should be renamed as core_visualization_shader_minimal.py.

Tested on Linux, I still have to do the same on Windows

jf---:
Thanks for the feedback @tpaviot , this was more of a placeholder, but there really is no reason not to merge now... 

Hopefully more soon regarding `GLSL` soon... its a pretty exciting field...

tpaviot:
Yes I will merge as oon as:

* a new line is added a the end of the srcipt ;

* the shape is not a sphere anymore. We should load the cylinder_head.brep file, the sphere is IMO not the suitable shape to test shaders/lights etc. 

* the test works on Windows too (still be done).

jf---:
hi @tpaviot , this all makes sense, though I'm not sure what with regard to Windows involves here?

tpaviot:
@jf--- on Windows:

```
  File "core_visualization_glsl_antialias.py", line 156, in <module>
    display.Context.Redisplay(anIO.GetHandle())
RuntimeError: Standard_Failure
CSF_ShadersDirectory or CASROOT is set incorrectly
```
An additional env var must be set

tpaviot:
@jf--- In this pr, we should add more than one example. I really don't understand glsl and its requirements. I don't even get the occ standard shaders work (the phong one, raytrace). Coloring a shape in green using shaders is absolutely useless, the minimal shader example is ok but another one must come with it, bringing a spectacular visual effect.

jf---:
hi @tpaviot , yes, we should. This PR is more to discuss progress regarding the topic.

 so I've made some progress porting the pink.(vs/fs) shader that you used in a previous webgl example. 
doing so requires being able to update the `time` shader variable.
currently, some methods are missing that are required:

```python
aProgram.Variables()
# non-wrapped pointer, same object
# Out[2]: <Swig Object of type 'NCollection_Sequence< Handle_Graphic3d_ShaderVariable > *' at 0x1158bc330>

# missing method, possible due to usage of templates?
# aProgram.PushVariable(TCollection_AsciiString("time), time_+=1)


```

Here's the reference to the [`Graphic3d_Shader`](https://www.opencascade.com/doc/occt-6.9.0/refman/html/class_graphic3d___shader_program.html#a597d040ae2e5882b410867ff80b3e66e) docs, for additional reference.

That said, for now, making this a `Qt` specific example would allow to side-step the missing methods.
That's the approach explored in the [chinese blog post](http://www.cppblog.com/eryar/archive/2015/07/21/211331.html) mentioned before.

So, ideally for this example something like a `ambient occlusion` shader would be implemented

tpaviot:
@jf--- can you plese open 2 issues for the missing methods ?

jf---:
Done. 

[Three.js](https://threejs.org/examples/#webgl_materials_variations_physical) provides some interesting ideas wrt shader examples...

tpaviot:
@jf--- yep, but I don't think these are *GLSL* shaders. Copy/pasting shaders from threejs to oce doesn't work (at least I did not get it working)

jf---:
@tpaviot , you recall the pink shader in your webgl example, right? so I'm close to porting that.
it requires some remapping of variable names, you need to use those in `Declarations.glsl` rather than those used in `Three.js`...

here's a snippet of `Declarations.glsl`

```glsl

// Vertex attributes
#ifdef VERTEX_SHADER
  attribute vec4 occVertex;
  attribute vec3 occNormal;
  attribute vec4 occTexCoord;
  attribute vec4 occVertColor;
#endif

// Matrix state
uniform mat4 occWorldViewMatrix;  //!< World-view  matrix
uniform mat4 occProjectionMatrix; //!< Projection  matrix
uniform mat4 occModelWorldMatrix; //!< Model-world matrix

uniform mat4 occWorldViewMatrixInverse;    //!< Inverse of the world-view  matrix
uniform mat4 occProjectionMatrixInverse;   //!< Inverse of the projection  matrix
uniform mat4 occModelWorldMatrixInverse;   //!< Inverse of the model-world matrix

uniform mat4 occWorldViewMatrixTranspose;  //!< Transpose of the world-view  matrix
uniform mat4 occProjectionMatrixTranspose; //!< Transpose of the projection  matrix
uniform mat4 occModelWorldMatrixTranspose; //!< Transpose of the model-world matrix

uniform mat4 occWorldViewMatrixInverseTranspose;  //!< Transpose of the inverse of the world-view  matrix
uniform mat4 occProjectionMatrixInverseTranspose; //!< Transpose of the inverse of the projection  matrix
uniform mat4 occModelWorldMatrixInverseTranspose; //!< Transpose of the inverse of the model-world matrix
```

jf---:
@tpaviot , thanks for the feedback wrt setting `CASROOT` on windows.
it probably makes most sense to have this as a utility function somewhere.

setting `CASROOT` is not so well defined - if the `conda` version is used, then probably yes, since the directory layout is known...

the function could look like:

```python
def set_casroot_conda():
    try:
        casroot = os.environ["CASROOT"]
    except KeyError:
        share_dir = os.path.abspath(os.path.join(OCC.__path__[0], "..", "..", "..", "..", "share")) 
        oce_dirs = glob.glob(os.path.join(share_dir, "oce-*"))
        nr_oce_dirs = len(oce_dirs)
        if nr_oce_dirs == 1:
            os.environ['CASROOT'] = oce_dirs[0]
        elif nr_oce_dirs == 0:
             raise EnvironmentError("no OCE directory found, where expected: \n{}".format(share_dir))
        else nr_oce_dirs:
             raise EnvironmentError("multiple OCE directories found, where expected: \n{}".format(share_dir))
```

tpaviot:
@jf--- this is not a big issue. I manually set up this env variable to enable the script on Windows. It works as on Linux (I can see a green sphere). It's cool if you can port the pink shader example.

jf---:
OK, thanks for now...

PR #375, opened 2016-12-22
resizing issue of the qtViewer widget fixed #359  the methode `resizeGL` calls a non-existing method in Qt5, best to remove it

jf---:
resizing issue of the qtViewer widget fixed #359
 the methode `resizeGL` calls a non-existing method in Qt5, best to remove it

### DIFF
--- a/src/Display/qtDisplay.py
+++ b/src/Display/qtDisplay.py
@@ -95,6 +95,26 @@ class qtBaseViewer(QtOpenGL.QGLWidget):
             self._display.OnResize()
 
 
+def handle_retina(_nsview_pointer):
+    screen = QtGui.QGuiApplication.instance().primaryScreen()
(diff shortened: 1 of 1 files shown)


tpaviot:
@jf--- there are two different things in this PR removal of the resizeGL method that causes issue with Qt4, and something related to the retina display.

Regarding the retina code, using objc is a workaround, I think this shoud be done at the c++ level, in Display3d.cpp. Since the beginning, pythonocc is built on swig only to interact with c or objc objects. Adding pyobjc or ctypes code is not consistent.

tpaviot:
commit rebased, objc dependency added to conda receipe see commits 11f940055a9c42b29d7b665d20d4de350678a4b7 and 5654ace2b09583d845e27ac5f3823c78ab260111

PR #377, opened 2016-12-22
try to throw an exception wrt OpenLocalContext on Travis

jf---:
try to throw an exception on Travis ( times out one mine )


tpaviot:
@jf--- exception is not thrown

jf---:
@tpaviot , correct, also I checked earlier OCE version, where I had no problems.
I'll rebuild things to investigate further locally

PR #411, opened 2017-04-01
Added GEOMAlgo_Splitter

trelau:
The source files are in src/Splitter for now. The *.i files were created using the pythonocc-generator just pointed to the GEOMAlgo_Splitter.hxx file. The two source files are from Salome 8.2.0 release. You'll notice I copied the BOPAlgo_MakeVolume.hxx directly into the GEOMAlgo_headers.i file because the "#include <BOPAlgo_MakerVolume.lxx>" line was causing SWIG compile issues. It is commented out. A simple example is also provided.


tpaviot:
@trelau Great, thanks.

PR #434, opened 2017-06-20
Load STEP with colors

hinxx:
@tpaviot here is cleaned up version of the code that will load STEP file and apply colors by reading color definitions from the STEP file itself. It is very noisy with console output, feel free to comment those print()'s out.


tpaviot:
@hinxx it's just perfect, thanks for sharing. I ran a few tests, seems to work, I will report any issue

hinxx:
Glad to be able to contribute. 

I did run into some minor issues when trying it out with our huge STEP files with ten thousands of labels. Some shapes would appear in the wrong place. No idea on how to debug it though. I'll look into it more closely as time permits.

PR #458, opened 2017-10-05
_build/ directory should be ignored generally.

tommy3001:
_build/  for documentation can be in the sub folder (e.g. doc/examples/_build/ ).


tommy3001:
Errors during make html documentation. Small alignments in documentation

PR #466, opened 2017-11-06
remove and ignore .DS_Store file

thorade:
https://gist.github.com/octocat/9257657
https://github.com/github/gitignore/blob/master/Python.gitignore


tpaviot:
okay

PR #468, opened 2017-11-19
Extended api documentation for the module Addons

tommy3001:
I added some rst code to the Addons.i file to extend the apidoc for this module. 
But some of the existing name definitions are seeming to be a little bit unclear. 
For example the Module name says not really something about the provided functionality. 
The parameters are starting partially with "the", for example theName, theX1 or theLayer without any obvious reference.


tpaviot:
@tommy3001 That's a welcome addition, thanks for your contribution. This also could be done in Visualization.i if ever you have time

tommy3001:
ShouOk. I will take a look in the next days.

PR #483, opened 2017-12-24
Text alignments in BRepPrimAPI documentation

tommy3001:
Minor text alignments in BRepPrimAPI documentation.

Merry Xmas.


tpaviot:
@tommy3001 we don't use to modify .i files directly. Indeed, they are automatically created from the occt c++ headers files, using a python script. This project is known as pythonocc-generator (see https://github.com/tpaviot/pythonocc-generator).

The function that constructs the module docstring is here: https://github.com/tpaviot/pythonocc-generator/blob/master/src/generate_wrapper.py#L662

If you wish to modify the module docstring, it then should be done at the generator function (that way, all .i files will e modified accordingly)

tpaviot:
and the function docstring can be modified at https://github.com/tpaviot/pythonocc-generator/blob/master/src/generate_wrapper.py#L715

tommy3001:
Thanks for the advice! I will check it.

PR #486, opened 2017-12-27
Option to render edges

aothms:
* Option to render edges
* Sync to pythreejs 1.0.0b3

This is not really intended to be merged, just a quick fyi on my progress. It seems that backwards compatibility with the 0.4 branch would require a substantial amount of extra code (not just the couple of lines I added), so we need to see what version(s) you are planning to target in the end.

notebook_edges

Things I am still planning to work on:
* Navigation and better initial camera rotation center point
* Interactivity (clickable surface and callback fn)
* Default jupyter repr function on TopoDS_Shape

### DIFF
--- a/src/Display/WebGl/jupyter_renderer.py
+++ b/src/Display/WebGl/jupyter_renderer.py
@@ -23,26 +23,55 @@ import tempfile
 
 try:
     from pythreejs import *
-    from IPython.display import display
+    from IPython.display import display, HTML
     from ipywidgets import HTML, Text
     from traitlets import link, dlink
     import numpy as np
 except ImportError:
     raise AssertionError("You must install pythreejs and dependencies to benefit from the jupyter renderer")
 
+from OCC.Bnd import Bnd_Box
+from OCC.BRepBndLib import brepbndlib_Add
 from OCC.Visualization import Tesselator
 
-default_shape_material = PhongMaterial(color='#a6a6a6', shininess=0.9, morphTargets=True)
+# default values
+default_shape_color = '#%02x%02x%02x' % (166, 166, 166)
+
+
+def get_boundingbox(shape, tol=1e-5):
+    """ return the bounding box of the TopoDS_Shape `shape`
+    Parameters
+    ----------
+    shape : TopoDS_Shape or a subclass such as TopoDS_Face
+        the shape to compute the bounding box from
+    tol: float
+        tolerance of the computed boundingbox
+    use_mesh : bool
+        a flag that tells whether or not the shape has first to be meshed before the bbox
+        computation. This produces more accurate results
+    """
+    bbox = Bnd_Box()
+    bbox.SetGap(tol)
+    brepbndlib_Add(shape, bbox, False)
+
+    x_min, y_min, z_min, x_max, y_max, z_max = bbox.Get()
+    x_size = x_max - x_min
+    y_size = y_max - y_min
+    z_size = z_max - z_min
+    x_center = x_min + x_size / 2
+    y_center = y_min + y_size / 2
+    z_center = z_min + z_size / 2
+    return x_size, y_size, z_size, x_center, y_center, z_center
+
 
 class JupyterRenderer(object):
-    def __init__(self):
-        self._camera = PerspectiveCamera(position=[30, 30, 30],
-                                         fov=20,
-                                         children=[DirectionalLight(color='#ffffff', position=[50, 50, 50], intensity=0.5)])
-        self._background = 'black'
+    def __init__(self, size='100%%'):
+        self._background = 'white'
         self._background_opacity = 1
-        
-    def DisplayShape(self, shp, shp_material = default_shape_material):
+        self._size = size
+
+
+    def DisplayShape(self, shp, shape_color = default_shape_color):
         """ takes a topods_shape and displays it in a jupyter notebook
         """
         # compute the tesselation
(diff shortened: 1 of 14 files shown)


src/Display/WebGl/jupyter_renderer.py
```diff
@@ -87,35 +98,59 @@ def DisplayShape(self, shp, shape_color = default_shape_color):
...
+        })
+        
         #shape_geometry.exec_three_obj_method('computeFaceNormals') TODO normals should be client-side computed
         #shape_geometry.exec_three_obj_method('computeVertexNormals') TODO idem for vertex normals
-        shp_material = PhongMaterial(color=default_shape_color, shininess=0.9, morphTargets=True)
+        shp_material = MeshPhongMaterial(color=default_shape_color, shininess=0.9, morphTargets=True)
         shape_mesh = Mesh(geometry=shape_geometry, material = shp_material)
-        scene_shp = Scene(children=[shape_mesh, AmbientLight(color='#101010')])
 
+        edges = list(map(lambda i_edge: [tess.GetEdgeVertex(i_edge, i_vert) for i_vert in range(tess.ObjEdgeGetVertexCount(i_edge))], range(tess.ObjGetEdgeCount())))
```
aothms: Oops, this still needs to go under the `if render_edges:` block

tpaviot:
Excellent. Backward compatibility with 0.4 is not a requirement, I will drop 0.4 support as soon as I get 1.0.0b3 working ! You can thus remove all 0.4 only code.

By the way, did you do something special to install pythreejs1.0.0b3 ? "special" in  the sense "not documented on the pythreejs page"

tpaviot:
@aothms I confirm that the target is pythreejs 1.0.0.bxxx.

aothms:
> By the way, did you do something special to install pythreejs1.0.0b3 ? "special" in the sense "not documented on the pythreejs page"

I am working on this on a jupyterhub docker spawner ubuntu image. In case it might help you can find the relevant part below. It took me some fiddling with users and paths [1].

[1] https://github.com/jovyan/pythreejs/pull/150

~~~dockerfile
WORKDIR /opt/build
RUN git clone https://github.com/jovyan/pythreejs
WORKDIR /opt/build/pythreejs
RUN chown -R jovyan .
USER jovyan
RUN /opt/conda/bin/pip install --user -e .
WORKDIR /opt/build/pythreejs/js
RUN npm run autogen
RUN npm run build:all
USER root
RUN jupyter nbextension install --py --symlink --sys-prefix pythreejs
RUN jupyter nbextension enable pythreejs --py --sys-prefix
~~~

Unfortunately I have not yet succeeded in making selection work [2] (which was one of the main use cases for me for migrating to pythreejs) so that's a bit of a bummer.

[2] https://github.com/jovyan/pythreejs/issues/152

I have worked a bit on multiple shapes in the same viewport, I added kwarg `show` and `Display()` method for that. What do you think? Perhaps `display(j)` would also be an option, `AddShape()`?

~~~py
j = jupyter_renderer.JupyterRenderer()
j.DisplayShape(box1, render_edges=True, show=False)
j.DisplayShape(box2, render_edges=True, shape_color=(1.,0.,0.), edge_color=(0.,1.,0.), show=False)
j.Display()
~~~

tpaviot:
Selection work: I use to work with 3js and BufferGeometries for a while, selection works perfectly. Run the core_webgl_threejs_* examples from the latest pythonocc-0.18.1 release, and check the related code.

Attached a zip archive including all files generated from the torus examples. You have a basic selection/event implemented.
[toruses_pyocc0181_3jsr88.zip](https://github.com/tpaviot/pythonocc-core/files/1591633/toruses_pyocc0181_3jsr88.zip)

The selection mechanism also required to build a mapping between a buffergeometry Id and the related topods_shape.

Adding shapes: to be consistent with GUI based renderer, something like this would be cool:

renderer.DisplayShape(shape1)
renderer.DisplayShape(shape2)
renderer.start_display()

shape1 and shape2 would be added to a shapelist to be renderer.

tpaviot:
I finally scceeded to install the current dev version. I studied the conda-forge recipe and extracted the following line:
```python setup.py install --single-version-externally-managed --record record.txt ```
and then install/enable the jupyter extension.
Tested on my windows 10 machine in a dedicated conda environment

tpaviot:
I merged your first commit (option to render edges, amended see 9104870bab31d3f5b65e6ab626e6bc9d4c4f1b91) and bump to 1.0.0x version. It works as expected. Moving up to pythreejs 1.0.0b3, I noticed a few regressions:

* the 3d window is much smaller. I don't know how to set the widget initial size ;

* the quality seems to be worse (a lot of aliasing) ;

* the zoom, using the middle mouse rolling button, does not work as expected.

did you ?

aothms:
> I finally scceeded to install the current dev version

Wonderful, I removed the few backwards comp lines

> ... and extracted the following line:

Magic at work

> the 3d window is much smaller. I don't know how to set the widget initial size

> the quality seems to be worse (a lot of aliasing) ;

I have a width, height on the Renderer (https://github.com/tpaviot/pythonocc-core/pull/486/commits/46253d84a8f7136e572f5ab42545ae2514dc29d3#diff-b509ee812727b2a79c36a29c0479574fR255). Maybe that's different from your approach of size (perhaps that simply scales the pixels but leaves the opengl viewport the same, I didn't have a look).

> the zoom, using the middle mouse rolling button, does not work as expected.

I see we arrived at the same conclusion and used OrbitControls :)

---

I have selection in a reasonably working state now, but you need my PR in pythreejs as well: https://github.com/jovyan/pythreejs/pull/153 at the very least the first commit to deal with missing UVs. In my code the edge linesegments are no longer candidates for picking so the 2nd commit there is not strictly necessary.

aothms:
There is also the option of `antialias = True` on the `Renderer` btw.

tpaviot:
I also noticed the width, height and antialias parameters. I picked-up your first commit (render edges), and commited a patch over it. you can rebase your work on top of review/tess and submit another PR

tpaviot:
@aothms The latest commit of review/tess is sync with your progresses, thank you. I modified a bit the selection mechanism. This works quite well with the changes you pushed to the pythreejs repos regarding mesh picking.

aothms:
Cool, thanks for getting things in sync

tpaviot:
I close this PR, everything has been merged into review/tess (will soon merge into master). Please submit another PR if you make progresses.

PR #506, opened 2018-02-06
Fix conda install command.

ikeough:
I found when attempting to install pythonocc via conda that the install command in `README.md` was not correct. Fixing this caused a successful install.


tpaviot:
Thanks, I missed it.

jkbgbr:
while we are at it: the line above the corrected conda command is also incorrect: it simply installs in the active (conda) environment.
(sorry, idk how to create a pull request)

tpaviot:
@jkbgbr Fixed in current master.

PR #507, opened 2018-02-07
Review/structure refactor

tpaviot:
New project structure: modules moved to OCC.Core, examples moved to pythonocc-demos, doc moved to pythonocc-documentation


jf---:
A huge step @tpaviot!
I think the repo split is a great idea.
These days, most python project support a `requirements.txt` file specifying dependencies.
Question is whether the `pythonocc-utils` should be part of this list?
I'd be much in favor of doing so, since it would help defining which modules are close to the core, and which ones are more peripheral.

tpaviot:
The requirements.txt is part of a pip process as far as I know, whether the project rather rely on conda. But I don't think it's necessary to add pythonocc-utils as a dependency. Indeed, as you underline, some classes/functions are very close to the Core (for instance traverse a topology, compute a bounding box, read a STEP file etc.) Actually, some of these important classes/functions has been moved back to the pythonocc-core project, in the Extend package. The structure is currently the following:
OCC
--> Core : all modules that directly wrap an oce package (for instance gp, TopoDS etc.). These python modules are generated using SWIG
--> Display : 100% pure modules related to visualization (Qt and wx Gui, Webgl, jupyter notebook etc)
--> Extend : 100% pure python modules built upon the Core, that provide shortcuts for most common operations, i.e. functions that you'll necessary have to develop/use if you go with pythonocc for any project.

You can see that all examples (https://github.com/tpaviot/pythonocc-examples) have been ported to this new structure.

This way, I do think that t's much more clear, and it will be easier to write the related documentation.

jf---:
I like the `OCC.Core` idea, making it much more obvious whether this is pure python or at the wrapper level. Great idea!

PR #528, opened 2018-04-07
fix transparent shape rendering in jupyter_renderer.py

LinJiarui:
add missed parameter 'opacity' when initializing MeshPhongMaterial


tpaviot:
+1

PR #577, opened 2018-11-12
Remove `calling_script` from image filename

aaronang:
# Before

Without this patch, the generated image filenames can be pretty weird. For example:

```
/local/home/aang/Code/hello/hello/test/capture-/local/home/aang/Applications/pycharm-2018-1-1542064678.jpeg
``` 

> Note: This happens when using PyCharm.

# After

With this patch, the generated image filenames will look like:

```
/local/home/aang/Code/hello/hello/test/capture-1-1542064946.jpeg
```


tpaviot:
@aaronang you're right. The calling_script variable can be safely removed since it's not used anymore

PR #578, opened 2018-11-13
Use `update` parameter in `OffscreenRenderer.DisplayShape`

aaronang:
When `update=True` it will always resize the view to the shapes, which is not always desired.


tpaviot:
Good, thanks

aaronang:
@tpaviot Just curious. How frequent is the package updated on Anaconda Cloud?

PR #583, opened 2018-11-18
Review/tp transparent handles

### LINKED ISSUE #539: Idea: Make Handle_* classes completely transparent

tpaviot:
Follows PR #581 from @rainman110 

The OCCT handle smart pointer API is a C++ specific detail and should not be exposed to Python. Instead it should be completely invisible. This PR almost completely removes Handles_ from pythonOCC.

This makes the use of pythonOCC much cleaner and less verbose!

I used the std::shared_ptr swig library as a basis for the implementation.
To be compatible with most of the "old" code, GetHandle() and GetObject() methods still exist but do nothing except from a deprecation warning. In general, I tried to make sure that the code is backwards compatible as possible.

I adapted already the unit tests to not include the legacy code. The demos are working unmodified but should be adapted to avoid deprecation warnings.

The only thing that is not backwards compatible is, when previously an Null-Handle was returned - e.g. a non successful computation. Before, the object was a handle but .IsNull() was true. Now, it is a None type, which is more Pythonic.

The "magic" is implemeted in new swig typemaps in OccHandle.i.

Closes #539

Examples have been refactored accordingly see https://github.com/tpaviot/pythonocc-demos


tpaviot:
Merging, thanks @rainman110

PR #691, opened 2019-10-30
WIP: Implemented support for occt 720

### LINKED ISSUE #576: OpenCASCADE 7.2.0 support
I have seen, there is a branch to support OpenCASCADE 7.2.0. What is the current status?

Do you need any help?

rainman110:
This is tested only on linux at the moment. The
heart of the new implementation is a new handle
wrapping. Before, all transients were wrapped with
a handle in python. This lead however to problems in the
inheritance chain now. To fix this, all objects are now wrapped
without the handle, but I use swigs reference counting mechanism
to make sure that everything is correct memory-wise.

Other changes are mainly to make the compilation and tests work.
This includes:
 - Fixed order of includes (In OCCT 720, some headers are not correct
   on their own but required pre-including of e.g. Interface_TypedValue.hxx)
 - Delayed definition of template NCollection based containers.
   In order to make the type system work correctly, transients have to be wrap_handle'd
   first before defining arrays on them.
 - Removal of NCollection_Handle<>
 - Added import of Geom.i in TColGeom.i. This was required, to make the tests work.

How did I test:
 - First test of the handle wrapping was performed on a standalone code. Here,
   I made sure, that valgrinds shows no errors or memory leaks. This makes me
   pretty comfortable, that the implementation is solid.
 - I tested the inheritance hierarchy with and without inheritance from standard_transient.
 - I made sure that all types Handle_XXX, Handle(XXX) and opencascade:handle<XXX> all work the same.
 - I tested all tests from run_tests.py. Here two tests are failing:
    1. The number of edges seems to have changed in one tests. I did not have a look into it.
       I suppose this os more OCCT 720 related than the python wrapping.
    2. One test is failing that does not corretly return the enum BRepCheck_Status. This was pretty tough
       and must be a problem of the NCollection_List parsing of swig. I stopped trying for now
       but this should be addressed later!

Still to be done:
 - Testing on windows and macOS
 - Testing with all demos
 - Put all required changes into the generator

Fixes #576


tpaviot:
Testes on Ubuntu 18.04, gcc-7.4, swig 3.0.12. Compilation/installation works fine. 88 tests out of 90 pass. 2 failures :

````
File "/home/thomas/Devel/pythonocc-core/test/core_wrapper_features_unittest.py", line 539, in test_return_enum
    self.assertEqual(los1.First(), BRepCheck_Multiple3DCurve)
AssertionError: <Swig Object of type 'BRepCheck_Status *' at 0x7f811f27de10> != 5
````

and

````
  File "/home/thomas/Devel/pythonocc-core/test/core_extend_topology_unittest.py", line 126, in test_vertex_edge
    self.assertEqual(len(edges_from_vert), topo.number_of_edges_from_vertex(vert))
AssertionError: 3 != 6
````

The first one is related to the wrapper itself. I remember fighting this BRepCheck_Status issue for a long time before I give up.

The second one is related to the TopologyExplorer class implementation, it certainly has to be updated to occt720.

rainman110:
These are also the two mentioned failing tests I referred to. the first issue is seemi to be pretty hard. I spent a day on it, but I stopped for now. 
It seems to be a swig issue parsing NCollection_List correctly. My suggestion here is, two first enable all warnings . Maybe we'll get the idea then what's going wrong.

The second is an occt detail not related to the wrapper (probably)

tpaviot:
Maybe it's worth moving up to the latest swig release (4.0.0). I did not tested it yet, but it's more than 2 years of development over the previous release. I guess many parsing issues have been fixed.

tpaviot:
So if I understand correctly, the order in the wrapper files, before the classes definition, is:

1 enums
2 wrap_handle*
3 template*
4 typedefs*

right ?

I will change the generator to fit your changes, and rebase all of this on top of the master branch. Then, the next step is, in my opinion, to target occt-7.4.0 (it was released a few days ago, 7.2.0 is two years old).

rainman110:
I suppose yes. Going for OCCt 7.4.0 is definitely the way to go!

rainman110:
One more thing: I am trying to make the windows build working. OCCT decided for backwards compatibility, that in some cases on windows, Handle_XXX != Handle(XXX)

Instead the Handle_XXX classes are derived from Handle(XXX). This is an important detail. In OCCT 7.4, this is however not the case anymore.

What has to be done for OCCT 7.2.0?
Do not use Handle_XXX in the swig input files. This can make problems. Always prefer Handle(XXX) or opencascade::handle<XXX> (which is the same). The generator should take care of this.

One particular example, which made problems:
https://github.com/tpaviot/pythonocc-core/blob/ec3cc043e6ef85b2c46ca315008aa23fd980399d/src/SWIG_files/wrapper/BOPDS.i#L56

Here, Handle_XXX is used explicitly.

Another problematic line was:
https://github.com/tpaviot/pythonocc-core/blob/ec3cc043e6ef85b2c46ca315008aa23fd980399d/src/SWIG_files/wrapper/BRepBuilderAPI.i#L592

@tpaviot Could you adapt the generator to not use Handle_ anymore?

tpaviot:
@rainman110 ok for the use of Handle(XXX) everywhere in replacement for Handle_XXX.

rainman110:
@tpaviot Just saw, that the generator explicitly transforms Handle(XXX) into Handle_XXX, starting from here:

https://github.com/tpaviot/pythonocc-generator/blob/e955f4b532b1b433226f82796c4cf75271560803/src/generate_wrapper.py#L601

So the block just have to be removed, right?

tpaviot:
@rainman110 I don't if it simply has to be removed, or be remained and replaced back, at the end, by Handle(XX). To be honest, I don't remember exactly why I had to replace Handle(XXX) to Handle_XXX. According to the comment, it has to do with Cppheaderparser, I have to check that.

rainman110:
@tpaviot If Cppheaderparser makes a problem, you can simply also use opencascade::handle\<xxx\> instead.

rainman110:
@tpaviot This transformation is indeed done before parsing the header files. So using the template based opencascade::handle<XXX> should be safe.

jf---:
A long time ago I looked into using the LLVM c++ parser ( which has python bindings ) rather than Cppheaderparser. It might be a bit more robust. It was pretty easy to swap out these components of memory serves. Is this still of interest?

rainman110:
I think @trelau also is using a clang based parser here: https://github.com/LaughlinResearch/pyOCCT_binder

tpaviot:
The problem does not come from the header parser, CppHeaderParser is quite good, only one or two headers out of thousands fail.  The issue is rather the SWIG interface file generation from the result of the parser, which is independant from any parser. I used to work with gccxml at the beginning of the project, then moved to cppheaderparser. The good thing is that it's pure python, lightweight, fast.

tpaviot:
@rainman110 I renamed wip/occt7x-support to review/occt7x-support (but without removing the old one). I'm updating this branch from your changes (the new OccHandle) and a few changes related to occt7.4.0 support. I'm currently working on an update to the generator that includes the opencascade::handle change. I will update all the interface files when ready to be tested.

rainman110:
@tpaviot , @jf--- Next progress... successfully build it on windows. It was a bit tricky with many undefined referenced in the end.

rainman110:
@tpaviot I fixed the windows builds

Here's a short summary for you to integrate into the generator:
 - Updated OccHandle
 - Removed methods Aspect_DisplayConnection (const TCollection_AsciiString & theDisplayName), GetAtom, GetDisplay, GetDisplayName from class Aspect_DisplayConnection 
 - Removed FillInitialVectorOfSolution  from class BRepApprox_TheImpPrmSvSurfacesOfApprox 
 - Removed class BRepClass3d_BndBoxTreeSelectorLine  and class BRepClass3d_BndBoxTreeSelectorPoint 
 - Removed GetSafeOffset from class BRepOffset_MakeSimpleOffset
 - Removed FillInitialVectorOfSolution  from class GeomInt_TheImpPrmSvSurfacesOfWLApprox
 - Put include for gp_XYZ and gp_XY in front of all NCollection_CellFilter include (again bug of opencascade)

The Handle_XXX to Handle(XXX) is not part of this commit. It would be too many changes to see, whats really going on.

tpaviot:
@rainman110 I won't work on occtè20 anymore. You can leave this PR opened if you with to continue the discussion on this topic, otherwise let's go on with occt740 support

rainman110:
We can close this, but we should keep some of the notes. They might come handy when going for Windows

Issue #751, opened 2020-01-05
Option for rendering edges in JupyterRenderer as continuous in addition to dashed


Krande:
Hello,

I came across some issues using the JupyterRenderer to render a simple wire (comprised of 2 edges) using the following code (and pythnocc 7.4.0RC1)

```
from OCC.Core.gp import gp_Pnt
from OCC.Extend.ShapeFactory import make_wire
from OCC.Core.BRepBuilderAPI import BRepBuilderAPI_MakeEdge

h = 0.18
wbtn = 0.033
p1 = (0, 0, 1)
p2 = (0, -h, 1)
p3 = (wbtn, -h, 1)


e1 = BRepBuilderAPI_MakeEdge(gp_Pnt(*p1), gp_Pnt(*p2)).Edge()
e2 = BRepBuilderAPI_MakeEdge(gp_Pnt(*p2), gp_Pnt(*p3)).Edge()

mywire = make_wire([e1, e2])

# Visualize using SimpleGui
from OCC.Display.SimpleGui import init_display
display, start_display, add_menu, add_function_to_menu = init_display()
display.EraseAll()
display.DisplayShape(mywire)
display.FitAll()
start_display()

# Visualize using JupyterRenderer
my_renderer = JupyterRenderer(size=(800, 600))
my_renderer.DisplayShape(mywire, update=False, render_edges=False)
my_renderer.Display()
```

This is visualized just fine using the SimpleGui in the code above

![image](https://user-images.githubusercontent.com/9642232/71782647-e33da500-2fdc-11ea-94f5-4271c2bea8a5.png)

However the JupyterRenderer resulted displaying only one of the edges as shown below:

![image](https://user-images.githubusercontent.com/9642232/71782694-56dfb200-2fdd-11ea-99a3-350959f77374.png)

Now the problem (as far as I can tell) is the dash-spacing in the default dashed line geometry in the JupyterRenderer. By editing the above example using larger numbers (as shown below) I am able to display both edges.

```
from OCC.Core.gp import gp_Pnt
from OCC.Extend.ShapeFactory import make_wire
from OCC.Core.BRepBuilderAPI import BRepBuilderAPI_MakeEdge

h = 100
wbtn = 10
p1 = (0, 0, 1)
p2 = (0, -h, 1)
p3 = (wbtn, -h, 1)


e1 = BRepBuilderAPI_MakeEdge(gp_Pnt(*p1), gp_Pnt(*p2)).Edge()
[...]

PR #753, opened 2020-01-06
Replaced LineSegment with Line to eliminate dashed line behaviour

Krande:
As discussed in issue #751 here is the proposed solution using Line opposed to LineSegments


tpaviot:
Thank you for the report and your contribution.

PR #785, opened 2020-03-06
TopologyUtils: fix misspelling

karl-nilsson:
I ran into this bug recently.


tpaviot:
Good catch, thank you

PR #803, opened 2020-04-01
Add a flask server for WebGL (Threejs).

RobinCPC:
Wrap ThreejsRenderer to export occ shape as json data object(but not save to files).
Wrap html related class into a individual index.html
Utilize Flask jinja2 template to load json data and dynamical create javescript script to load shapes and Edges.
Update renderer to display vertices (point cloud) in WebGL.


tpaviot:
great, thanks

PR #844, opened 2020-05-27
getters, self._, face colors

andreasplesch:
small fixes


andreasplesch:
https://github.com/tpaviot/pythonocc-core/pull/844/commits/b857f6d36d88fc1cb3500926cacdc5c3c46613dc adds logic to distinguish between cases where 
a) subshapes have uniform face colors (Buggy suspension model)
b) subshapes have multiple colors (Kuka model)
Only in case a) the supershape, the shape containing the subshapes is added to the scenegraph. In case b) the supershape is not added because it leads to repeated geometries.

PR #851, opened 2020-06-08
Fixed global change of log level

jnwalther:
Importing the init_display function e.g. using from `OCC.Display.SimpleGui import init_display` will globally change the log level to DEBUG. Aside from this being an undesirable side effect during import, it also causes cluttered output when running larger test suites involving pythonocc. Therefore I propose the attached change, which causes the log level to be set only for the local logger instead of globally.

### DIFF
--- a/src/Display/backend.py
+++ b/src/Display/backend.py
@@ -1,5 +1,4 @@
 import logging
-import sys
 
 # backend constants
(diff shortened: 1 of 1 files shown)


tpaviot:
good point

PR #963, opened 2021-02-11
Add debian repo for packaging

Trophime:
Added debian directory for building package 
Actually desinged for 7.4.0


tpaviot:
:+1:  This is a great contribution !

tpaviot:
@Trophime did you try with 7.4.1 ? That should work without the BOPAlgo and missing step file patches.

PR #992, opened 2021-04-13
Mesh data source

tpaviot:
Since the support for SMESH was dropped, pythonocc lacks a way to display STL files as meshes. This branch adds a related module.


tpaviot:
An example was uploaded here https://github.com/tpaviot/pythonocc-demos/blob/master/examples/core_mesh_data_source.py

tpaviot:
New example https://github.com/tpaviot/pythonocc-demos/blob/master/examples/core_mesh_data_source_color_map.py

PR #997, opened 2021-05-04
a way to get a return from th GetSelectedShape function

Tanneguydv:
by return the first element of the selected_shapes list


tpaviot:
Interesting. However I'm not in favor of storing a mapping between ais_shapes and topods_shapes. Indeed this can be done outside the OCCViewer module, from your qt5app (the DisplayShape methods returns the ais_shapes).

But you're right there's an issue with the GetSelectedShape method, and it should be clear that this method returns an AIS_Shape, not a TopoDS_Shape. I'll have a look at it

Tanneguydv:
Thank you for your reply @tpaviot .
Sorry I mixed my commit files and a «test» one is first commit, the real changes are in the second one with :
     -   return self.selected_shape
     +   return self.selected_shapes[0]

I didn't catch that the expected return of the function was an ais shape though

tpaviot:
The DisplayShape method actually returns a list of AIS_Shapes. If you pass one TopoDS_Shape to the DisplayShape method, you get a list of only one AIS_Shape.

Tanneguydv:
Ok sorry I didn't see there was a link between Getselectedshape and displayshape!

PR #1006, opened 2021-05-19
MeshDS numpy interface

kleinsimon:
I added two constructors to Mesh_DS, one allowing the direct instantiation by 2d c++ arrays and one for pointer arrays of vertices and faces.
The first one is not correctly matched by SWIG, so I removed it from the swig wrapper file. But using the second one, one can now instantiate the MeshDS using two numpy arrays, which removes the need for multiple function calls for every element and vertex, resulting in a huge speedup.
I tested the interface using a few meshes, which gave the same result as using the original method.
For compilation, the numpy headers must be available, so maybe this is just a build-option.


tpaviot:
Superseded by #1062

PR #1042, opened 2021-10-22
Fix bug in pickling TopoDS_Shape: orientation is not conserved

cafhach:
Please note

- this PR incorporates a commit proposed in PR #1040 
- The changes in this PR are caused by changes in pythononcc-generator, see reference below

Pickling of TopoDS_Shape instances now uses BRepTools::Write & BRepTools::Read, which - different to BRepTools_ShapeSet - conserves the object's orientation. In order to be able to use BRepTools::Write/Read I added two wrappers WriteToString/ReadFromString which reside in the breptools class and have an alias in the BRepTools 
module breptools_WriteToString and breptools_ReadFromString


tpaviot:
superseeded by #1060

PR #1043, opened 2021-10-22
Pickling does not round

cafhach:
Up until now pickling rounded the shape. In our application this lead to slight differences between calculations done in the same process and calculations done in a subprocess (which was communicated with via pickle/unpickle). Building on top of PR #1042 I added an option to breptools_WriteToString to have full precision. In this case then the writing is done with a stringstream using 17 digits and a scientific floating point format. 

Perhaps - this would be a design decision up to you - it's more reasonable to always use full precision in breptools_WriteToString. I already made `full_precision = True` the default. My idea was that someone might still want to be able to exactly reproduce previously written brep files. 

Please note that 

- this PR is based upon previously commited PRs #1040 and #1042
- this PR contains the actual change in the generated code files and is logically connected to PR tpaviot/pythonocc-generator#83


tpaviot:
superseeded by #1060

PR #1049, opened 2021-11-04
Adding a layer management tool to pythonocc

Tanneguydv:
I propose an example file in another pull request (pythonocc-demos) along with this one

### DIFF
--- a/src/Extend/layer_mgmt.py
+++ b/src/Extend/layer_mgmt.py
@@ -53,10 +53,10 @@ class Layer:
         for shape in self.list_to_display:
             self.display.Context.Erase(shape, False)
 
-    def join(self, layer, clear=False):
+    def merge(self, layer, clear=False):
         """
-        :param layer: name of the layer to join
-        :param clear: bool to clear the layer to join
+        :param layer: name of the layer to merge to the main one
+        :param clear: bool to clear the layer 
         :return: None
         """
         for shape in layer.get_shapes():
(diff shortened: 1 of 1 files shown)


tpaviot:
Very good idea. The ```create``` and ```add``` methods look very similar. I suggest that the ```create``` method is removed, add a parameter to the ```__init__``` and call ```add``` from it. That could look like:

```python
class Layer:  # no need for parenthesis

    def __init__(self, from_shape=None, color=0):
        self.clear()
        if shape is not None:
            self.add(shape, color)
```

rainman110:
@Tanneguydv Currently, your code assumes ``display`` as a global variable, which might not always be the case. I suggest to add a display argument to the layer constructor.

Tanneguydv:
@tpaviot Indeed, the two functions were redundants, But I prefer to keep the color argument in the init section so that each layer has its  own color

@rainman110 True, I was thinking of this as most of the time display = self.display

Thank you,
here what I suggest :
```
class Layer:

    def __init__(self, from_display, shape=None, color=0):
        self.clear()
        self.color = color
        self.display = from_display
        if shape is not None:
            self.add(shape)

    def show(self):
        for shape in self.list_to_display:
            self.display.Context.Display(shape, True)

    def hide(self):
        for shape in self.list_to_display:
            self.display.Context.Erase(shape, False)

    def add(self, shape):
        self.to_display = {self.count}
        self.to_display = self.display.DisplayShape(shape, color=self.color)[0]
        self.list_to_display.append(self.to_display)
        self.count += 1
        self.display.Context.Erase(self.to_display, False)

    def clear(self):
        self.list_to_display = []
        self.count = 0
        self.to_display = {self.count}
```

if ok, I would change my pull request and adapt the example file in the pythonocc-demos repo

Tanneguydv:
I added two methods to the pull request : join layers, and get_shapes from layer

PR #1056, opened 2021-11-19
Optimizing topology exploration

igormaurell:
Using HashCode function to increase the performance of __loop_topo_ and __map_shapes_and_ancestors_ functions. To overcome the problem of a possible same HashCode for slightly different shapes, the method _IsSame_ is used as a confirmation step. In my tests, this solution optimize methods like _.faces()_ or _.edges()_ by more than 100x.

I ask you to test if the solution is complete and useful for the library. Any changes that need to be made, I am available.

Thank you.

### DIFF
--- a/src/Extend/TopologyUtils.py
+++ b/src/Extend/TopologyUtils.py
@@ -139,6 +139,7 @@ class TopologyExplorer:
         for face in srf.faces:
             processFace(face)
         '''
+        MAX_INT = 2**31 - 1
         topoTypes = {TopAbs_VERTEX: TopoDS_Vertex,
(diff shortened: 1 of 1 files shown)


tpaviot:
:+1: That's a great contribution. This code is more than ten years old and was never improved.

jf---:
Very cool, performance was never addressed or even gauged. Wonderful contribution. IIRC the class is reasonable well tested and your implementation careful. So of it doesn't break the test suite, it seems reasonable to merge.

igormaurell:
That is great. PythonOCC is helping a lot in my current projects. If I find some other possible contributions to make, I will let you know through other PRs. Thank you.

PR #1060, opened 2021-11-20
Fix pickling see https://github.com/tpaviot/pythonocc-generator/pull/83

tpaviot:


cafhach:
Thank you for review and merge :-)

PR #1142, opened 2022-09-11
Port to 7.6.3

tpaviot:
occt763 is binary compliant with occt762, the generator cant detect any API change

FYI @looooo


looooo:
so should we loosen the pinning in the conda-forge recipe?

tpaviot:
@looooo I still have some changes to merge, I will ping you when it is ready, thx!

PR #1190, opened 2023-03-24
wrap OCCT-7.7.1 (branch CR33261)

tnakaicode:
I tried to wrap on OpenCASCADE branch CR33261 on Win10 (Windows only). 3rd-party used the library that was in the 7.7.0 Installer.

The output of pythonocc-genrator as it was had the following three problems, so I fixed code by manual and built it.

1. pythonocc-generator output adds ```class uniton{}`` to Graphic3d_TransformPers in Graphic3d.i and causes a Build Error.
2. in pythonocc-generator output, StepRepr and StepShape cylclic import, which causes an Error when Python is executed.
3. the module that imports XCAFPrs, XCFDoc, ... Packege causes DLL not found when executing Python.
   3rd issue is most likely a problem during OpenCASCADE build in my enviroment, and may not be reproduced in other environments.


tpaviot:
I already built occt-771 and the related wrapper on my local machine, however, we still have to wait for occt-771 to be available on conda-forge, and let the azure build pass. I suggest you run the generator on linux, I didn't try on windows, I'm not sure the output is the same

tnakaicode:
In your Linux environment, did you have any problems importing after Wrapped?
In my Windows environment, after installing in the Python environment, I got error "DLL not found" in the following modules.
Since it was not OCCT-7.7.0, it seems to be an error specific to OCCT-7.7.1.

- _BRepBlend
- _BRepExtrema
- _BSplCLib
- _GeomInt
- _StdPrs
- _STEPConstruct
- _XCAFDoc
- _XCAFPrs

tpaviot:
occt-771 binaries have been pushed to conda-forge a few hours ago (https://anaconda.org/conda-forge/occt). Porting pythonocc-core up to 771 can be started using azure ci workflows

tpaviot:
@tnakaicode can you please rebase your branch on master? I updated the current master with the wrapper for occt771. Im' not sure about the workflow you used,  I'd like to see what's different in your pr

PR #1234, opened 2023-07-19
Support wayland

ovo-Tim:
If `$XDG_SESSION_TYPE` is `wayland`, run `export QT_QPA_PLATFORM=xcb`  to make program run on X11.


tpaviot:
This doesn't work on my ubuntu-22.04 machine:
```bash
qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.

Available platform plugins are: vkkhrdisplay, xcb, minimalegl, offscreen, wayland-egl, wayland, vnc, minimal, eglfs, linuxfb.

Abandon (core dumped)
```
(running PyQt6 from pip install)

Issue #627, opened 2019-02-11
Successfull compilation but libTKernel.so.11: cannot open shared object file on import


calocedrus:
I am installing python-occ on a linux box (ubuntu 18.04) and in a virtualenvironment(*). 

- Compiled and installed successfully OCE (in a folder under my /home directory).
- Compiled and installed (*apparently*) successfully python-occ the following way:
  - create a virtual environment for the install
  - cloned the latest branch with `git clone `
  - from a build directory, run `cmake` with the following options, specifying the location for the "non standard" location of OCE:
`cmake -DCMAKE_INSTALL_PREFIX=/home/ldv/.virtualenvs/cad_eg -DOCE_INCLUDE_PATH=/home/ldv/libs/OCE/include/oce -DOCE_LIB_PATH=/home/ldv/libs/OCE/lib ..`
  - run `make` (it takes a long time but compilation reports no error)
  - run `make install` (no need for `sudo` since the install target is in my user directory), all seems fine, with the last few lines from the install being:
 ```
[...]
-- Installing: /home/ldv/.virtualenvs/cad_eg/lib/python3.6/site-packages/OCC/XmlMXCAFDoc.py
-- Installing: /home/ldv/.virtualenvs/cad_eg/lib/python3.6/site-packages/OCC/XmlObjMgt.py
-- Installing: /home/ldv/.virtualenvs/cad_eg/lib/python3.6/site-packages/OCC/XmlTObjDrivers.py
-- Installing: /home/ldv/.virtualenvs/cad_eg/lib/python3.6/site-packages/OCC/XmlXCAFDrivers.py
-- Installing: /home/ldv/.virtualenvs/cad_eg/lib/python3.6/site-packages/OCC/gce.py
-- Installing: /home/ldv/.virtualenvs/cad_eg/lib/python3.6/site-packages/OCC/gp.py
-- Installing: /home/ldv/.virtualenvs/cad_eg/lib/python3.6/site-packages/OCC/math.py
```
In a python console I can do `import OCC`, no error.

But then when I try to import for example `OCC.IGESControl` from the same python console (with the virtual environment activated of course), I get:
```
File "/home/ludovic/.virtualenvs/cad_ge/lib/python3.6/site-packages/OCC/Core/IGESControl.py", line 25, in swig_import_helper
    return importlib.import_module(mname)
  File "/home/ludovic/.virtualenvs/cad_ge/lib/python3.6/importlib/__init__.py", line 126, in import_module
[...]

PR #1293, opened 2023-12-17
Update INSTALL.md

onrdmr:
I followed instruction in install.md but something are missing and found this issue so added missing part of installation on INSTALL.md. Following the issue  https://github.com/tpaviot/pythonocc-core/issues/627


tpaviot:
:+1:

Issue #1096, opened 2022-04-01
Read from stream in DataExchange


Fairyon:
Hi everyone!
I get a filestream of STEP and IGES files from a REST API and want to analyse it. I saw, that STEPControl_Reader supports reading from stream, but the DataExchange class does not.

I have wrote my own extension read_step_stream(datastream, ...), where I replaced just the line of read_step_file(filename, ...)
`status = step_reader.ReadFile(filename)`
with this one:
`status = step_reader.ReadStream(datastream)`
and the same for iges

Would it be possible for you to add function read_step_stream(datastream, ...) in a further releases?

Thanks in advance.

Alex aka Fairyon

cafhach:
Hi Fairyon,

great work. The next step would be to create a pull request containing your extension for @tpaviot to oversee and eventually accept.

JohannesVerherstraeten:
Hi, I have the same use case. And I believe [this issue](https://github.com/tpaviot/pythonocc-core/issues/1009) is about the same too.

It would be great if this feature can be added!

## Problem statement
To be clear, instead of loading my STEP file like this (the default way):
```python
def read_step_file(file_path: str) -> TopoDS_Shape: 
    step_reader = STEPControl_Reader()
    step_reader.ReadFile(file_path)
    step_reader.TransferRoot()
    return step_reader.Shape()
shape = read_step_file("my_file.step")
```
I'd like to be able to load it like this: 
```python
def read_step_file(file_stream: io.BytesIO) -> TopoDS_Shape: 
    step_reader = STEPControl_Reader()
    step_reader.ReadStream(file_stream)     # This method exists, but doesn't work with python streams. See below
    step_reader.TransferRoot()
    return step_reader.Shape()

# Here we read the stream from file, but in reality it would be a HTTP request body. 
with open("my_file.step", "rb") as f:
    stream = io.BytesIO(f.read())

shape = read_step_file(stream)
```
## Possible hint towards a solution?
I noticed that the `OCC.Core.STEPControl.STEPControl_Reader` class inherits `OCC.Core.XSControl.XSControl_Reader`, which has a `ReadStream` method, but this method doesn't accept python streams. 
```python
class XSControl_Reader(object):
    ...
    def ReadStream(self, *args) -> "IFSelect_ReturnStatus":
        r"""
        Loads a file from stream and returns the read status.

        Parameters
        ----------
        theName: char *
        theIStream: std::istream
        """
[...]

PR #1299, opened 2024-02-08
Allow reading/writing step files to string

anuraaga:
In #1096 there is great interest in being able to marshal step files without temporary files on disk. While `*Stream` methods are provided, they are awkward to use with Python files, and I think in some places in the library this is acknowledged and methods that read and write strings are provided. Strings do require buffering so may not be ideal compared to a stream in all cases, but in many cases it actually already is buffered, and it is still a significant improvement over temporary files for many of the cases listed in the issue, it would be great to be able to use this without files.

I mostly copied the pattern of breptools

https://github.com/tpaviot/pythonocc-core/blob/master/src/SWIG_files/wrapper/BRepTools.i#L855

If there's a reason to avoid this, no worries in closing this as it's out of the blue. But since the change wasn't big I figured I'd give it a shot.


test/core_ocaf_unittest.py
```diff
@@ -119,6 +119,67 @@ def test_read_step_file(self) -> None:
         a_shape = shape_tool.GetShape(label_shp)
         self.assertFalse(a_shape.IsNull())
 
+    def test_read_write_step_string(self) -> None:
+        """Writes a step file as a string and reads it back"""
```
anuraaga, line 123: I thought that doing a round trip would be the simplest way to ensure both writing and reading have assertions applied to them. I noticed that `test_write_step_file` may not have any assertions being applied

tpaviot:
Thank you for this interesting contribution.

However, the thing is you still need to have the step file stored on your local filesystem, and I can't see the improvement compared to the standard `Read` method. It would be much better that the `ReadFromString` method only takes a string, let's say:

```python
with open("test.stp", "r", encoding="uft8") as step_file:
    step_file_content = step_file.read()
step_reader.ReadFromString(step_file_content)
```
This way, the `ReadFromString` method could take any string, coming from instance from an API call.

tpaviot:
Good point ! If the filename is not necessary, let's just remove it (pass an empty string as filename), we don't need it at the python level (I have to check however why the filename is a mandatory parameter).

FYI I dont modify swig files directly in the wrapper directory, I use to work with the pythonocc-generator:

* I commit/push changes to the pythonocc-generator

* I generate all the swig files into the pythonocc-core /wrapper directory

* I commit changes to pythonocc-core and monitor the ci process

tpaviot:
Using an empty string seems to work, but there maybe some side effects I can not see so far.

tpaviot:
Reverse engineering of the `ReadStream` method of `STEPCAFControl_Reader`:

* this method is part of the `STEPCAFControl_Reader` class (see https://github.com/Open-Cascade-SAS/OCCT/blob/master/src/STEPCAFControl/STEPCAFControl_Reader.cxx#L366).

It simply calls the `ReadStream` method of the `myReader` instance with type `STEPControl_Reader`

* in the `STEPControl_Reader` class, the `ReadStream` method is defined here: https://github.com/Open-Cascade-SAS/OCCT/blob/master/src/STEPControl/STEPControl_Reader.cxx#L171

The stream name parameter is used at line https://github.com/Open-Cascade-SAS/OCCT/blob/master/src/STEPControl/STEPControl_Reader.cxx#L201 :

```c++
WS()->SetLoadedFile(theName);
```
The `WS` instance is a `XSControl_WorkSession`

The stream is read by calling the method `ReadStream` of the `aLibrary` instance (line 187) with type `IFSelect_WorkLibrary` defined here: https://github.com/Open-Cascade-SAS/OCCT/blob/master/src/IFSelect/IFSelect_WorkLibrary.cxx#L91

It just returns `1`. Damned, dead end.

* the `XSControl_Session` inherits the `SetLoadedFilename` method from the upper `IFSelect_WorkSession` class, see https://github.com/Open-Cascade-SAS/OCCT/blob/20955d88da8615c54bd9f4221a3ff9fbe90fd915/src/IFSelect/IFSelect_WorkSession.hxx#L150

```c++
 void SetLoadedFile (const Standard_CString theFileName)
  { theloaded = theFileName; }
  
  //! Returns the filename used to load current model
  //! empty if unknown
  Standard_CString LoadedFile() const
  { return theloaded.ToCString(); }
```

* there are only 10 occurrences of the `LoadedFile` method all over the occt codebase. As far as I understand, it is mostly used in the `DRAW` framework, whici is not covered by pythonocc.

***Conclusion*** I guess we can safely use an hardcoded string for the stream name parameter. We can wrap it as an optional parameter in python, if ever someone needs to set it to a different value.

anuraaga:
Thanks for confirming! In that case I lean towards not exposing a parameter, and if for some reason someone asks for it it could be added with a default value to be backwards compatible (I think). I somehow expect for no one to ever ask for it.

> pythonocc-generator

Sorry I hadn't noticed this repo. Should I move this change to it? And also

- Remove name parameter
- ReadFromString / WriteToString? I went with matching the grammar of `ReadStream` for this PR but noticed that pattern in the codebase too, happy to use either.

tpaviot:
I've been working on a more robust solution : wrap all std::istream and std::ostream parameters in occt to python strings. This leads to a simpler generator, and a python wrapper more aligned with occt method names and signature. It maybe less pythonic, but easier to go from the official documentation to python. PR is on the way

tpaviot:
@anuraaga see #1301

anuraaga:
Thanks! That looks very useful

hunosdof:
I am trying to load a step file from a stream instead directly from a file.

However, the following function fails:

`status = step_reader.ReadStream(file_path, file_content)`

I am passing this file_content:
        if isinstance(file_content, bytes):
            file_content = file_content.decode('utf-8')

It is in utf-8 format.
However, I get this error:

pyqt6 backend - Qt version 6.7.3
####### 3D rendering pipe initialisation #####
Display3d class initialization starting ...
Aspect_DisplayConnection created.
OpenGl_GraphicDriver created.
V3d_Viewer created.
AIS_InteractiveContext created.
V3d_View created
Graphic3d_Camera created
Graphic3d_StructureManager created
WNT window created.
Display3d class successfully initialized.
#########################################
OpenGl information:
  GLvendor: Intel
  GLdevice: Intel(R) RaptorLake-S Mobile Graphics Controller
  GLversion: 4.6.0 - Build 32.0.101.5763
  GLSLversion: 4.60 - Build 32.0.101.5763
  Max texture size: 16384
  Max FBO dump size: 4096x16384
  Max combined texture units: 192
  Max MSAA samples: 16
  Viewport: 1279x959
  Window buffer: RGB8 ALPHA8 DEPTH24 STENCIL8
  ResolutionRatio: 1
  FBO buffer: GL_SRGB8_ALPHA8 GL_DEPTH24_STENCIL8

I assume that the file_content argument is incorrect (but str is expected).
Is there a specific format this string has to be in?

Thank you!

PR #1301, opened 2024-02-22
Add wrapper support for file streams 

tpaviot:
Superseeds #1299 

This branch adds support for `Standard_IStream` and `Standard_OStream` as python ***strings***

c++ methods:

``` c++
ReadStream(std::istream & theIStream)
```
are wrapped as
```python
ReadStream("some python string")
```

and

``` c++
WriteStream(std::ostream & theOStream)
```
can be used in python with:

``` python
some_string = WriteStream()
```

This is used for object serialization (as an ascii string `Dump` or `DumpJson`  `InitFromJson`) as well as import export (STEP or GLTF files).

This is achieved using the set of `IOStream.i`  swig templates, and a change over the whole interface files generated from https://github.com/tpaviot/pythonocc-generator/commit/d168c34703748b98df1eaf66369be789dba2f180


tpaviot:
Note: pyi files must be updated

PR #1311, opened 2024-04-08
enhance the display url for all addresses 0.0.0.0 while using httpd

EeyoreLee:
The original log `## Open your webbrowser at the URL: http://localhost:42525` is ambiguous when I already set `addr` to `0.0.0.0`. It looks like the server runing only on localhost.
After merge this PR, when addr isn't `0.0.0.0`, the log will show as 
```
## Open your webbrowser at the URL: http://localhost:42525
```
and show as the following when addr is `0.0.0.0`
```
## Running on all addresses (0.0.0.0)
## Open your webbrowser at the URL: http://192.168.33.111:42525
```

### DIFF
--- a/src/Display/WebGl/simple_server.py
+++ b/src/Display/WebGl/simple_server.py
@@ -58,7 +58,7 @@ def get_interface_ip(family: socket.AddressFamily) -> str:
     :meta private:
     """
     # arbitrary private address
-    host = "fd31:f903:5ab5:1::1" if family == socket.AF_INET6 else "10.253.155.219"
+    host = "2001:db8::1" if family == socket.AF_INET6 else "192.0.2.1"
 
     with socket.socket(family, socket.SOCK_DGRAM) as s:
         try:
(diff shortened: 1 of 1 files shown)


EeyoreLee:
@tpaviot - Hi, the function `get_interface_ip` is inspired of [werkzeug#L669](https://github.com/pallets/werkzeug/blob/6f461e9485514821bd249d356147c3bb8b4e8697/src/werkzeug/serving.py#L669). Counld we trust it directly and skip the security of this part.

tpaviot:
@EeyoreLee I see an hardcoded IP address "10.253.155.219". Can you please review sourcery.ai suggestions

EeyoreLee:
@tpaviot - There  is a choice for server backend(`flask` & `http`). If users already installed `flask`, this code snipet will also execute indeed right now. `werkzeug` is a necessary dependency for [flask](https://github.com/pallets/flask/blob/2c7f57ad5bd19743312eb34ea5fb838a1349474e/pyproject.toml#L24). So I think align the ip address for original code will be more safe. This's just my own point. If you got it and still perfer sourcery.ai suggestion, I'll follow it.

PR #1365, opened 2024-08-26
Fix CMake check for NOTFOUND

atismer:
The check in CMake for the not found header does not work with my CMake Version 3.30.2. I think this check is also not correct for other CMake Versions.

## Summary by Sourcery

Correct the CMake condition for checking the presence of the 'gp_Pnt.hxx' header file to ensure compatibility with various CMake versions.

Bug Fixes:
- Fix the CMake check for the presence of the 'gp_Pnt.hxx' header file to correctly handle cases where the file is not found.


atismer:
@sourcery-ai review

PR #1373, opened 2024-10-10
Update DataExchange.py read_step

rsm-gh:
When using as_compound=True, the method was returning a shape instead of a compound, if there was only 1 shape to convert.

## Summary by Sourcery

Correct the return behavior of the read_step_file function to ensure it returns a list of shapes when as_compound is True, even if there is only one shape.

Bug Fixes:
- Fix the behavior of the read_step_file function to correctly return a list of shapes when as_compound is True and there is only one shape.


rsm-gh:
I included this commit because it makes more sens to always respect the same type of object, but actually when there is only 1 shape, I do not see much difference between the compound and the shape, so I guess that's why you removed it from newer versions (at 7.5 the code seems OK)

Issue #1350, opened 2024-07-10
String not interfaced to IStream


JJMinton:
Updating from 7.5.1 to 7.8.1, I am trying to replace `BRepTools_ShapeSet.ReadFromString` with `BRepTools_ShapeSet.Read` as I believe is the result of #1301 .

The resulting error was:
```python
>>> from OCC.Core.BRepTools import BRepTools_ShapeSet
>>> BRepTools_ShapeSet().Read("foo")
Traceback (most recent call last):
  File "<string>", line 1, in <module>
  File "/miniconda3/envs/occwl_dev/lib/python3.9/site-packages/OCC/Core/TopTools.py", line 2880, in Read
    return _TopTools.TopTools_ShapeSet_Read(self, *args)
TypeError: Wrong number or type of arguments for overloaded function 'TopTools_ShapeSet_Read'.
  Possible C/C++ prototypes are:
    TopTools_ShapeSet::Read(std::istream &,Message_ProgressRange const &)
    TopTools_ShapeSet::Read(TopoDS_Shape &,std::istream &)
```

I am installing from conda in a docker container:
```Dockerfile
# syntax=docker/dockerfile:1.4
FROM --platform=linux/amd64 python:3.9 as base

RUN python -m pip install --upgrade pip

FROM base as dev

# Insgtall OCC requirements
RUN apt-get update
RUN  apt-get install -y ffmpeg libsm6 libxext6

# Install conda
RUN mkdir -p ~/miniconda3 && \
    wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O ~/miniconda3/miniconda.sh && \
    bash ~/miniconda3/miniconda.sh -b -u -p /miniconda3 && \
    rm -rf ~/miniconda3/miniconda.sh && \
    /miniconda3/bin/conda init bash

ARG PROJ_DIR=/proj
ARG PIP_CACHE=/root/.cache/pip
WORKDIR ${PROJ_DIR}

# Create conda environment
COPY environment.yml .
RUN --mount=type=cache,target=/miniconda3/pkgs,id=conda \
    /miniconda3/bin/conda env create -f environment.yml
```
where `environment.yml` is
```yaml
name: occwl_dev
channels:
  - conda-forge
  - defaults
  - anaconda
dependencies:
  - python=3.9
  - numpy
  - pythonocc-core=7.8.1
  - networkx
  - wxPython
  - pydeprecate
```

[...]

PR #1384, opened 2024-11-12
Fix windows dll path

tpaviot:
This PR is a follow up for #1347

@farleyrunkel @Andrej730 @Bill-XU can you please review this PR I cannot test the fix.

## Summary by Sourcery

Fix the DLL path issue on Windows by adding a function to initialize OCCT libraries using the OCCT_ESSENTIALS_PATH environment variable. Update the INSTALL.md to provide a detailed build guide for Linux and Windows, including prerequisites, system requirements, and troubleshooting.

Bug Fixes:
- Fix DLL path issue on Windows by initializing OCCT libraries using the OCCT_ESSENTIALS_PATH environment variable.

Documentation:
- Revise and expand the INSTALL.md documentation to include a comprehensive build guide for both Linux and Windows, with detailed steps and troubleshooting tips.


Bill-XU (changes requested):
Thanks for the PR!

There is one miss after Ln.33.
A "break" is needed to prevent further processing the same directory.
The idea is call "add_dll_directory" upon a directory only if there is any DLL in it.

Bill-XU (changes requested):
Great THANKS ! It becomes more friendly to windows users.:-D

I have some suggestions.
- Regarding "System Requirements (Windows)" or "Installing Required Software (Windows)"
  rapidjson is also necessary for building RWGltf.
  Use `git clone https://github.com/Tencent/rapidjson.git`

- Regarding "Building pythonOCC (Windows) 3. Configure with CMake"
  If using CMake GUI, it is necessary not forgetting add two environment variables in GUI.
  `OCCT_INCLUDE_DIR=C:\OpenCASCADE-7.8.1\inc`
  `OCCT_LIBRARY_DIR=C:\OpenCASCADE-7.8.1\win64\vc14\lib`

- There is no explanation on how to configure rapidjson for RWGltf.
  Here is what I am doing.
  1. Run VisualStudio 2022 Community as administrator
  2. Open PYTHONOCC.sln generated by CMake
  3. Open properties window of project RWGltf
  4. Choose C/C++ on the left
  5. Edit "Additional Include Directories"
  6. Add include directory of rapidjson.
     Note: the include directory of rapidjson is the "include" itself under rapidjson root, and in it there is only one subfolder and its name is "rapidjson". In case you cloned rapidjson to C:\rapidjson, the include directory shall be "C:\rapidjson\include".
  7. If you continue building within VisualStudio, change target to Release, and build the whole solution then the project INSTALL.

  Unfortunately, I only use VisualStudio 2022 Community, so don't know how to make the same changes when using cmake --build.

- Regarding the value of "OCCT_ESSENTIALS_PATH", I suggest using the root directory of OpenCascade and its 3rd party libraries. (Maybe changing the name to "OCCT_ESSENTIALS_ROOT" is better?)
  In my case, I created a folder named "OpenCASCADE-7.8.1" under "C:\Program Files (x86)". And I extracted all 3rd party libraries in it, also created a folder named "opencascade-7.8.1" for holding compiled OCCT. So I set the value to "C:\Program Files (x86)\OpenCASCADE-7.8.1".
  *This folder structure is also recommended by OCCT in this instruction https://dev.opencascade.org/doc/overview/html/index.html#intro_install_windows*
  (If the value was "%CASROOT%\win64\vc14\bin", the 3rd party libraries, like TCL/TK DLLs,  would not be loaded when the folder structure was as the same as OCCT recommended)

tpaviot:
@Bill-XU thank you for your comments. For anything related to documentation on building occt/pythonocc for windows, you can open a new issue/pr.

Andrej730:
I have a suggestion to include setup of `OCCT_ESSENTIALS_PATH` in cmake on windows, so it will be set up automatically during the pythonocc installation.

Andrej730:
> @Andrej730 You mean the `OCCT_ESSENTIALS_PATH` would be hard coded after the build is completed?

Yeah, I think, this is what most users would expect if build succeeded and Python package is installed - Python package to work without setting up additional environment variable manually.

tpaviot:
ok that makes sense I will add this feature

INSTALL.md
```diff
@@ -319,9 +330,9 @@ python3 examples/core_classic_occ_bottle.py
 ```bash
 >>> ImportError: DLL load failed while importing _gp:
 ```
-set the ```OCCT_ESSENTIALS_PATH``` environment variable
+set the ```OCCT_ESSENTIALS_ROOT``` environment variable
 ```batch
-setx OCCT_ESSENTIALS_PATH "%CASROOT%\win64\vc14\bin"
+setx OCCT_ESSENTIALS_ROOT "%CASROOT%\win64\vc14\bin"
```
Bill-XU, line 335: The "OCCT_ESSENTIALS_ROOT" must be the root of OpenCASCADE in which there are not only opencascade itself, but also its 3rd party libraries. Otherwise, many packages (AIS etc.) would throw DLL not found error.
The recommended folder structure is described in OCCT's documentation here  https://dev.opencascade.org/doc/overview/html/index.html#intro_install_windows

Bill-XU (changes requested):
Sorry, changing Ln.335 is not a suggestion but a correction.
Please check my comment, thank you!

tpaviot:
I'm sorry @Bill-XU I don't get the point between the _ROOT, _PATHS, "root of opencascade" etc. this is specific windows stuff. You can let the OCCT_ESSENTIALS_ROOT env var point wherever needed. If you need a different behavior please submit a patch.

Bill-XU:
> I'm sorry @Bill-XU I don't get the point between the _ROOT, _PATHS, "root of opencascade" etc. this is specific windows stuff. You can let the OCCT_ESSENTIALS_ROOT env var point wherever needed. If you need a different behavior please submit a patch.

@tpaviot 
Err... my apologize for not saying it clearly.
The point is that, the value of the env var should not be "%CASROOT%\win64\vc14\bin", inside which there are only OpenCascade DLLs.
As @Andrej730 mentioned in #1347, it is necessary to load not only opencascade DLLs but also 3rd party libraries.

tpaviot:
@Andrej730 I tweaked CmakeList.txt so it becomes possible to define OCCT_ESSTNEIALS_ROOT at build time and hard code the related path to config.py. It can however still be defined at run time. Can you please test and report any issue?

@Bill-XU I think the INSTALL.md is now more accurate, can you please review.

tpaviot:
BTW @Andrej730 I remember you discussed a pythonocc issue, related to shape serialization and version upgrade, but I can't find it anymore, can you please point me to the right place?

Andrej730:
> I remember you discussed a pythonocc issue, related to shape serialization and version upgrade, but I can't find it anymore, can you please point me to the right place?

@tpaviot I believe this is the one - https://github.com/tpaviot/pythonocc-core/issues/1350#issuecomment-2227910142

PR #1386, opened 2024-11-13
Replaced deprecated method Poly_Triangulation Triangles()

rsm-gh:
Replaced deprecated method Poly_Triangulation Triangles() with Triangle()

## Summary by Sourcery

Enhancements:
- Replace deprecated method Poly_Triangulation::Triangles() with Triangle() to ensure compatibility with updated API.


tpaviot:
@rsm-gh good catch! Deprecation warnings are not raised on my machine, there's something I must have forgot to check/uncheck some option in cmake.

The ShapeTesselator code is quite old, it works but I guess it could benefit from a general cleanup. Please feel free to report any improvement.

PR #1396, opened 2024-12-08
Fixed numpy interface for surface derivative computation; added tests

jnwalther:
Hi, I was made aware of a bug in the python interfaces for the numpy surface derivative computation, which made me realize that there were no tests. This PR fixes the error and supplements the tests.

## Summary by Sourcery

Fix the numpy interface for surface derivative computation and add tests to validate curve and surface derivative evaluations.

Bug Fixes:
- Fix the numpy interface for surface derivative computation by correcting the function signature to include an additional parameter.

Tests:
- Add tests for curve and surface derivative evaluations to ensure correct functionality of the numpy interface.


tpaviot:
:+1:

PR #1484, opened 2026-03-25
fix: correctly set default PYTHONOCC_WRAP_VISU build option

justend29:
Previously, the `PYTHONOCC_WRAP_VISU` build setting was incorrectly set to a non-empty string instead of the value `OFF` [here](https://github.com/tpaviot/pythonocc-core/blob/master/CMakeLists.txt#L85). This caused all subsequent `if(PYTHONOCC_WRAP_VISU)` checks to evaluate the variable as set.

Furthermore, a user couldn't set `PYTHONOCC_WRAP_VISU` on the command line, as it was always overwritten within `CMakeLists.txt`.

## Summary by Sourcery

Fix configuration of the visualization wrapper build option to respect user overrides and correctly depend on OpenGL availability.

Bug Fixes:
- Ensure PYTHONOCC_WRAP_VISU defaults to OFF when OpenGL is not found instead of being set to a non-empty string that always evaluates as enabled.
- Allow PYTHONOCC_WRAP_VISU to be set from the CMake command line without being overridden in CMakeLists.

Build:
- Tie the default value of PYTHONOCC_WRAP_VISU directly to OPENGL_FOUND while centralizing OpenGL detection and warnings in the CMake configuration.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -70,20 +70,22 @@ endmacro(option_with_default OPTION_NAME OPTION_STRING OPTION_DEFAULT)
 #####################################################################
 # OpenGL. If available, enable compilation for Visualization module #
 #####################################################################
-find_package(OpenGL)
-include_directories(${OPENGL_INCLUDE_DIR})
+set(OPENGL_FOUND FALSE)
+if(NOT DEFINED PYTHONOCC_WRAP_VISU)
+  find_package(OpenGL)
+  include_directories(${OPENGL_INCLUDE_DIR})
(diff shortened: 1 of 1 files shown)


tpaviot:
Thanks. The non-empty string bug was fixed in the review/occt800 branch (OCCT 8 port), which also removed option_with_default, so this PR no longer applies. I made PYTHONOCC_WRAP_VISU a regular option defaulting to OPENGL_FOUND, so a value passed on the command line is kept, and asking for it without OpenGL is now an explicit configuration error instead of a silent OFF.

PR #1489, opened 2026-05-04
Fixed incorrect reported package version

### LINKED ISSUE #1488: Incorrect reported version in PkgBase/__init__.py
At the time of writing, pythonocc-core's latest version is 7.9.3, but the file 

https://github.com/tpaviot/pythonocc-core/blob/779838f352694711d60c1567ed3f6e06f3826aa2/src/PkgBase/__init__.py#L5-L8

reports 7.9.0.

This could be a problem, if other tools need the exact version, e.g. when they want to build extensions for pythonocc-core, that need exaxt ABI compatibility.

rainman110:
Fixes #1488

## Summary by Sourcery

Bug Fixes:
- Correct the package patch version constant from 0 to 3 so the reported version matches the released build.

### DIFF
--- a/src/PkgBase/__init__.py
+++ b/src/PkgBase/__init__.py
@@ -5,7 +5,7 @@ import platform
 # Version number
 PYTHONOCC_VERSION_MAJOR = 7
 PYTHONOCC_VERSION_MINOR = 9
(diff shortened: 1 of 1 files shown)


tpaviot:
:+1:

